PR #24, opened 2019-11-05
Added missing comments and omitempty tags for the ProcessArgs struct

PacoDw:
- Added missing comments
- Added omitempty tags for the ProcessArgs struct

### DIFF
--- a/mongodbatlas/clusters.go
+++ b/mongodbatlas/clusters.go
@@ -30,7 +30,7 @@ type ClustersServiceOp struct {
 
 var _ ClustersService = &ClustersServiceOp{}
 
-// AutoScaling configures your cluster to automatically scale its storage and cluster tier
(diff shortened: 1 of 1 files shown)


mongodbatlas/clusters.go
```diff
@@ -30,15 +30,18 @@ type ClustersServiceOp struct {
 
 var _ ClustersService = &ClustersServiceOp{}
 
+// AutoScaling configures your cluster to automatically scale its storage and cluster tier
```
themantissa: FYI We actually have two separate autoscaling features, compute is our new one (which we'll hopefully add soon tot he provider) and diskGBEnabled, which is here, is just for storage.  In order to reduce confusion I'd change the comment to : // AutoScaling configures your cluster to automatically scale its storage

PacoDw: okay, thank you ! 👍

PR #27, opened 2019-11-25
chore: added default disk size for each provider tier

PacoDw:
- Added default disk size map for each provider tier

It can be use as a following: 
`DefaultDiskSizeGB["AWS"]["M10"]`

This new feature is necessary to use it on the Terraform cluster resource to define the disk size automatically.

If you have a comment just let me know, thanks.

### DIFF
--- a/mongodbatlas/clusters.go
+++ b/mongodbatlas/clusters.go
@@ -142,7 +142,9 @@ var DefaultDiskSizeGB map[string]map[string]float64 = map[string]map[string]floa
 		"R200":      1500,
 		"M200_NVME": 3100,
 		"M300":      2000,
-		"M400":      3000,
+		"R300":      2000,
+		"R400":      3000,
+		"M400_NVME": 4000,
 	},
 	"GCP": map[string]float64{
 		"M10":  10,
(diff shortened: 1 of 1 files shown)


mongodbatlas/clusters.go
```diff
@@ -114,6 +114,59 @@ type clustersResponse struct {
...
+		"R60":       320,
+		"M60_NVME":  1600,
+		"M80":       750,
+		"R80":       750,
+		"M80_NVME":  1600,
+		"M140":      1000,
+		"M200":      1500,
+		"R200":      1500,
+		"M200_NVME": 3100,
+		"M300":      2000,
```
shum: I think we're missing `R300: 2000`

mongodbatlas/clusters.go
```diff
@@ -114,6 +114,59 @@ type clustersResponse struct {
...
+		"M400":      3000,
```
shum: `M400` is not a valid instance type. This should be `R400`

I think we're also missing `M400_NVME: 4000`.

mongodbatlas/clusters.go
PacoDw: Thank you so much, let me resolved 👍

mongodbatlas/clusters.go
PacoDw: thanks, it resolved it 👍

PR #33, opened 2019-12-09
Organization

PacoDw:
Added:
- Organization resource to handle the API calls
- Organization test cases


themantissa:
Organization work is on hold for now.

themantissa:
I am going to close this PR since as the full CRUD capabilities we need to support organizations doesn't yet exist.  Hence it will also be assumed for both the go client and any dependent work that the organization will have already been created.

PR #34, opened 2019-12-15
Improve Makefile

gssbzn:
- Check fmt using golangci
- Proerly install golangci
- Fix error when doing `make build`


themantissa:
@gssbzn - thank you for the submission!  I will have the devs take a look.

PR #38, opened 2019-12-27
Enable golint and fix errors

gssbzn:
This enables golint to be run as part of make lint and fixes some warnings flagged by it
this aims to make the codebase more homogenous as part of the code seemed to be linted and other not

**Warning**: this may be a breaking change as the name of some public interfaces are changing

### DIFF
--- a/mongodbatlas/clusters.go
+++ b/mongodbatlas/clusters.go
@@ -168,6 +168,7 @@ var DefaultDiskSizeGB map[string]map[string]float64 = map[string]map[string]floa
 		"M200": 256,
 	},
 }
+
 //List all clusters in the project associated to {GROUP-ID}.
 //See more: https://docs.atlas.mongodb.com/reference/api/clusters-get-all/
 func (s *ClustersServiceOp) List(ctx context.Context, groupID string, listOptions *ListOptions) ([]Cluster, *Response, error) {
--- a/mongodbatlas/clusters.go
+++ b/mongodbatlas/clusters.go
@@ -114,61 +114,6 @@ type clustersResponse struct {
 	TotalCount int       `json:"totalCount,omitempty"`
 }
 
-// DefaultDiskSizeGB represents the Tier and the default disk size for each one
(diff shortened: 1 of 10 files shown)


mongodbatlas/clusters.go
```diff
@@ -114,61 +114,6 @@ type clustersResponse struct {
 	TotalCount int       `json:"totalCount,omitempty"`
 }
 
-// DefaultDiskSizeGB represents the Tier and the default disk size for each one
-// it can be use like: DefaultDiskSizeGB["AWS"]["M10"]
-var DefaultDiskSizeGB map[string]map[string]float64 = map[string]map[string]float64{
```
gssbzn: this was not in use internally and there's no way for you to use it as `mongodbatlas. DefaultDiskSizeGB` so there's no point on keeping something like this
This is the easy approach but I'm happy to revert this or expose it as a `func`

themantissa:
Note: reaching out internally to @gssbzn.  Not assigning reviewers until then.

mongodbatlas/clusters.go
PacoDw: Could you remove it again please? it isn't already necessary in the Terraform part.

mongodbatlas/clusters.go
```diff
@@ -114,6 +114,60 @@ type clustersResponse struct {
 	TotalCount int       `json:"totalCount,omitempty"`
 }
 
+// DefaultDiskSizeGB represents the Tier and the default disk size for each one
+// it can be use like: DefaultDiskSizeGB["AWS"]["M10"]
+var DefaultDiskSizeGB map[string]map[string]float64 = map[string]map[string]float64{
```
PacoDw: Could you remove it again please? it isn't already necessary in the Terraform part.

gssbzn: Done

PacoDw: Thank you so much 👍

PR #43, opened 2020-01-07
Added the label attribute to the cluster struct

PacoDw:


mongodbatlas/clusters.go
```diff
@@ -77,6 +77,7 @@ type Cluster struct {
 	ClusterType              string                   `json:"clusterType,omitempty"`
 	DiskSizeGB               *float64                 `json:"diskSizeGB,omitempty"`
 	EncryptionAtRestProvider string                   `json:"encryptionAtRestProvider,omitempty"`
+	Labels                   []Label                  `json:"labels,omitempty"`
```
themantissa: struct? missing for clusters?

PacoDw: Related to the above comment, I needed the same Label struct to be used in the cluster as well. For this reason, I took the database users branch (which contains the struct label) to create this branch. Intending to use and recycle the same Label struct avoiding repeat the same code or waiting until the database PR was merged to start with this, so to work faster, I decided to do it like so. I'm so sorry the confusion.

themantissa: Ahhhh, that makes sense. Thank you for explaining.

PR #46, opened 2020-01-21
Simplify tests

gssbzn:
- Prefer deep.Equal over reflect.DeepEqual, some places were doing both some others a mix
- If there's an unexpected error then prefer t.Fatal over t.Error as Fatal will prevent any other case from executing (which would fail anyway)

### DIFF
--- a/mongodbatlas/alert_configurations_test.go
+++ b/mongodbatlas/alert_configurations_test.go
@@ -6,9 +6,8 @@ import (
 	"net/http"
 	"testing"
 
-	"github.com/mwielbut/pointy"
-
 	"github.com/go-test/deep"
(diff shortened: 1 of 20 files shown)


marinsalinas:
[Review Mongo Client PR #46](https://trello.com/c/fT1ALd6s/9-review-mongo-client-pr-46)

gssbzn:
@marinsalinas I have tagged @savannaendicott for an team review first, after that we should be good to other reviews, no rush here as this is mostly trying to simplify some of the test cases

themantissa:
@gssbzn thank you!  Meeting with the team today so catch them up on the process discussed (yesterday was a US holiday so wee behind).

PR #51, opened 2020-01-31
X509 auth db users

PacoDw:
Added:
- X509 to handle the API calls
- X509 test cases


marinsalinas:
Just solve the conflicts and we are ready to merge it.

PR #54, opened 2020-02-06
x509AuthDBUsers: Fixed the id type

PacoDw:
The id attribute eas changed to int type due that the MongoDB server returns this value with this type.


mongodbatlas/x509_authentication_database_users.go
```diff
@@ -34,7 +34,7 @@ type UserCertificate struct {
 	MonthsUntilExpiration int    `json:"monthsUntilExpiration,omitempty"` // A number of months that the created certificate is valid for before expiry, up to 24 months.default 3.
 	Certificate           string `json:"certificate,omitempty"`
 
-	ID        string `json:"_id,omitempty"`       // Serial number of this certificate.
+	ID        int    `json:"_id,omitempty"`       // Serial number of this certificate.
```
marinsalinas: Could you please change it to *int64  to avoid integer overflow errors

PacoDw: okay thanks 👍

PR #56, opened 2020-02-21
Fix incorrect json tag for groupId

gssbzn:
The proper key in the json is lower case, this was probably a typo or copy/paste error


gssbzn:
@themantissa happy to tag someone from my team but this is just one liner that should probably be addressed shortly

PR #57, opened 2020-03-04
Add Organization API calls

hieuln:
This PR adds the implementation for all Organization APIs as specified here https://docs.atlas.mongodb.com/reference/api/organizations/


themantissa:
@hieuln thank you for the submissions but we are not planning to add organizations to the Go client as this time for a few reasons.  May I ask what spawned your work and submission?   Are you using this go client yourself or wanting more organization support w/ in Terraform?

hieuln:
hi @themantissa,
We are trying to implement the automation of provisioning users/teams and assign permissions in Atlas from Okta. There are some limitations (or bugs?) from API that prevent us from completing this task. We can't add more users with our domain to Atlas via CreateAtlasUser API as API said this domain is reserved (?), so actually we need to know which users are already invited to the organization already in order to assign to a team (otherwise the assignment will fail)
So actually I only need the API call ListUsers within an Organization. With hundreds of devs and team switching, we can't do this manually, you know. Or do you have a better ideas to do it?

themantissa:
@hieuln I very much appreciate the information and situation you describe.   Is the core need currently to be able to know which users have been added to an org and accepted the invite so they can be added to a team?  

Second, I'll likely be closing this PR soon without merging since until some changes within the underlying Atlas system are fully rolled out we are not supporting the Organization or Atlas Users resources.  Do note that the newer connection string work is happening in another PR and will be added to Terraform in the near term.

hieuln:
Hi @themantissa ,
Yes, that's all my need. At the moment I can't get all users who are invited to the organization except via the Org API calls. If mongodb atlas will not support this API anymore, our automation will be broken. Is there another way to get this?

About the newer connection string, very appreciate that. In the mean time when the feature is not available yet, I forked the client and terraform provider and make a quick fix there so we can finish our setup. Of course we always want to use the upstream and don't want to maintain a local copy but the feature is not there yet. We will soon switch to the upstream. Actually it blocks us for the rollout at the moment.

PR #58, opened 2020-03-05
CLOUDP-57632: Implementing Alerts resource

andreaangiolillo:
[CLOUDP-57632](https://jira.mongodb.org/browse/CLOUDP-57632): 
Implement the Atlas API resource for alerts, https://docs.atlas.mongodb.com/reference/api/alerts/

### DIFF
--- a/mongodbatlas/atlas_alerts.go
+++ b/mongodbatlas/atlas_alerts.go
@@ -10,8 +10,8 @@ const alertPath = "groups/%s/alerts"
 
 type AlertsService interface {
 	List(context.Context, string, *ListOptions) ([]Alert, *Response, error)
-	GetAnAlert(context.Context, string, string) (*Alert, *Response, error)
-	AckAnAlert(context.Context, string, string, *AlertsRequest) (*Alert, *Response, error)
+	Get(context.Context, string, string) (*Alert, *Response, error)
+	Acknowledge(context.Context, string, string, *AcknowledgeRequest) (*Alert, *Response, error)
 }
 
 // AlertServiceOp handles communication with the Alert related methods
@@ -48,7 +48,7 @@ type Alert struct {
 }
 
 // AlertsRequest contains the request Body Parameters
-type AlertsRequest struct {
+type AcknowledgeRequest struct {
 	AcknowledgedUntil      string `json:"acknowledgedUntil,omitempty"`      // The date through which the alert has been acknowledged. Will not be present if the alert has never been acknowledged.
 	AcknowledgementComment string `json:"acknowledgementComment,omitempty"` // The comment left by the user who acknowledged the alert. Will not be present if the alert has never been acknowledged.
 }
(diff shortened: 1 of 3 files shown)


mongodbatlas/atlas_alerts.go
```diff
@@ -0,0 +1,155 @@
...
+
+// Alert represents MongoDB Alert.
+type Alert struct {
+	ID                     string           `json:"id,omitempty"`                     // Unique identifier.
+	GroupID                string           `json:"groupId,omitempty"`                // Unique identifier of the project that owns this alert configuration.
+	AlertConfigID          string           `json:"alertConfigId,omitempty"`          // ID of the alert configuration that triggered this alert.
+	EventTypeName          string           `json:"eventTypeName,omitempty"`          // The type of event that will trigger an alert.
+	Created                string           `json:"created,omitempty"`                // Timestamp in ISO 8601 date and time format in UTC when this alert was opened.
+	Updated                string           `json:"updated,omitempty"`                // Timestamp in ISO 8601 date and time format in UTC when this alert was last updated.
+	Enabled                *bool            `json:"enabled,omitempty"`                // If omitted, the configuration is disabled.
```
gssbzn: I'd rather not make this a pointer and not omit it as the behaviour would be the same but I don't have a strong opinion against this

mongodbatlas/atlas_alerts.go
```diff
@@ -0,0 +1,155 @@
...
+	CurrentValue           *CurrentValue    `json:"currentValue,omitempty"`           // CurrentValue represents current value of the metric that triggered the alert. Only present for alerts of type HOST_METRIC.
+	ReplicaSetName         string           `json:"replicaSetName,omitempty"`         // Name of the replica set. Only present for alerts of type HOST, HOST_METRIC, BACKUP, and REPLICA_SET.
+	ClusterName            string           `json:"clusterName,omitempty"`            // The name the cluster to which this alert applies. Only present for alerts of type BACKUP, REPLICA_SET, and CLUSTER.
+	Matchers               []Matcher        `json:"matchers,omitempty"`               // You can filter using the matchers array only when the EventTypeName specifies an event for a host, replica set, or sharded cluster.
+	MetricThreshold        *MetricThreshold `json:"metricThreshold,omitempty"`        // MetricThreshold  causes an alert to be triggered.
+	Notifications          []Notification   `json:"notifications,omitempty"`          // Notifications are sending when an alert condition is detected.
+}
+
+// AlertsRequest contains the request Body Parameters
+type AlertsRequest struct {
```
gssbzn: [nit] can we change this?
```suggestion
type AcknowledgeRequest struct {
```

mongodbatlas/atlas_alerts.go
andreaangiolillo: The alert_configurations is using a pointer for Enabled. If we decided to change this pointer, I think we should also change the one in alert_configurations.

gssbzn: Ok let's leave it as it is for the sake of continuity

mongodbatlas/atlas_alerts_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+import (
+	"fmt"
+	"github.com/go-test/deep"
+	"github.com/mwielbut/pointy"
+	"net/http"
+	"testing"
+)
+
+func TestAlert_GetAnAlert(t *testing.T) {
```
gssbzn: Your tests still reference the old name, let's keep them in sync please

gssbzn:
@themantissa @marinsalinas we updated this with latest master and should be ready for final review

PR #59, opened 2020-03-05
Make the ServiceOp depend on an interface

gssbzn:
Making the ServiceOp structs depend on an interfaces enables me to inject my own client for the Ops manager an Cloud manager API
This should be a non breaking change that enables a more flexible extention of the structs


gssbzn:
@themantissa ready for final review

gssbzn:
@andreaangiolillo while testing this in the ops manager client I found I actually need to make the client public, can you please give this another look

gssbzn:
Thanks @PacoDw

PR #63, opened 2020-03-12
Improve Travis

gssbzn:
These are just some minor improvements to travis to remove unused features and improve our coverage of go versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,18 @@
-dist: xenial
-sudo: required
-services:
-  - docker
 language: go
 
-git:
-  depth: 1
-
-install:
-- make tools
+go:
+  - 1.12.x
+  - 1.13.x
+  - 1.14.x
+  - tip
 
 matrix:
   fast_finish: true
   allow_failures:
     - go: tip
   include:
-  - go: "1.12.x"
-    name: "Code Lint"
+  - name: "Code Lint"
+    before_install: make tools
     script: make lint
-  - go: "1.12.x"
-    name: "Code UnitTest"
-    script: make test
+
+script: make test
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,23 +1,18 @@
-dist: xenial
-sudo: required
```
gssbzn: we don't need `sudo`

.travis.yml
```diff
@@ -1,23 +1,18 @@
...
-services:
-  - docker
```
gssbzn: we don't need docker, docker service only let's you run things like `docker pull` or `docker push` it doesn't mean that your test is running on docker

.travis.yml
```diff
@@ -1,23 +1,18 @@
...
 language: go
 
-git:
-  depth: 1
-
-install:
-- make tools
+go:
```
gssbzn: improve our coverage of go versions

.travis.yml
```diff
@@ -1,23 +1,18 @@
...
+  - tip
 
 matrix:
   fast_finish: true
   allow_failures:
     - go: tip
   include:
-  - go: "1.12.x"
-    name: "Code Lint"
+  - name: "Code Lint"
```
gssbzn: linting will only run once and as such we only need to install linting for this step

GNUmakefile
```diff
@@ -9,7 +9,7 @@ build:
 	go install ./$(PKG_NAME)
 
 test:
-	go test $(TEST) -timeout=30s -parallel=4 -cover
+	go test $(TEST) -timeout=30s -parallel=4 -cover -race
```
gssbzn: add a test for race conditions

PR #64, opened 2020-03-12
CLOUDP-58969: Create restore job returns a list and not a single job

gssbzn:
the documentation may be wrong and this is the actual response, I'm confirming this internally


gssbzn:
@themantissa I've confirmed this internally with the backup team so this is ready for review on your side, I'm also opening a ticket to update docs

PR #65, opened 2020-03-13
CLOUDP-58634: Add support for checkpoints in the Atlas client

andreaangiolillo:


mongodbatlas/atlas_checkpoints.go
```diff
@@ -0,0 +1,91 @@
...
+	"net/http"
+)
+
+const (
+	backupCheckpoints = "groups/%s/clusters/%s/backupCheckpoints"
+)
+
+// AtlasCheckpointService is an interface for interfacing with the Checkpoint
+// endpoints of the MongoDB Atlas API.
+type AtlasCheckpointService interface {
```
gssbzn: Can we remove the `Atlas` part? This is already the atlas client so it should be clear enoght already

mongodbatlas/continuous_snaphots.go
```diff
@@ -46,16 +46,19 @@ type ContinuousSnapshot struct {
...
+	EncryptionEnabled  bool              `json:"encryptionEnabled,omitempty"`
+	FileSizeBytes      int64             `json:"fileSizeBytes,omitempty"`
+	MasterKeyUUID      string            `json:"masterKeyUUID,omitempty"`
+	MongodVersion      string            `json:"mongodVersion,omitempty"`
+	ReplicaSetName     string            `json:"replicaSetName,omitempty"`
+	StorageSizeBytes   int64             `json:"storageSizeBytes,omitempty"`
+	TypeName           string            `json:"typeName"`
+	ShardName          string            `json:"shardName,omitempty"`
+	TokenDiscovered    bool              `json:"tokenDiscovered,omitempty"`
+	TokenTimestamp     SnapshotTimestamp `json:"tokenTimestamp,omitempty"`
```
gssbzn: Is you want to make this ommitempty you may need to make it a pointer

andreaangiolillo: Do you prefer making them pointers or creating a new struct? I am thinking about the latter

gssbzn: if this is the same for both but with extra fields let's make them pointers

gssbzn: another 3rd way is creating 3 structs, one base, one with only the extra and then one that compose both, composition sometimes is great ;)

mongodbatlas/mongodbatlas.go
```diff
@@ -61,7 +61,7 @@ type Client struct {
 	X509AuthDBUsers                  X509AuthDBUsersService
 	ContinuousSnapshots              ContinuousSnapshotsService
 	ContinuousRestoreJobs            ContinuousRestoreJobsService
-	AtlasCheckpoints                 AtlasCheckpointService
+	AtlasCheckpoints                 CheckpointService
```
gssbzn: missed the rename here, can we also make CheckpointService -> CheckpointsService?

mongodbatlas/checkpoints.go
```diff
@@ -0,0 +1,91 @@
...
+
+//CloudProviderSnapshotRestoreJobsServiceOp handles communication with the CloudProviderSnapshotRestoreJobs related methods of the
+//MongoDB Atlas API
+type CheckpointsServiceOp struct {
+	Client RequestDoer
+}
+
+var _ CheckpointService = &CheckpointsServiceOp{}
+
+type Checkpoint struct {
```
PacoDw: Thank you so much, it's really nice, but could you add comments describing all the importing functions and structs to have coherency? Also, it avoids warnings in some editors for missing comments.

PR #66, opened 2020-03-16
fix: added field roles in notifications for alert configurations, add…

coderGo93:
- Reformat code in alert configurations and its test.
- Added field `roles` in `notifications`
- Added field `roles` in tests for read, create and update

### DIFF
--- a/mongodbatlas/alert_configurations.go
+++ b/mongodbatlas/alert_configurations.go
@@ -76,28 +76,28 @@ type MetricThreshold struct {
 
 // Notification sends when an alert condition is detected.
 type Notification struct {
-	APIToken            string    `json:"apiToken,omitempty"`            // Slack API token or Bot token. Populated for the SLACK notifications type. If the token later becomes invalid, Atlas sends an email to the project owner and eventually removes the token.
-	ChannelName         string    `json:"channelName,omitempty"`         // Slack channel name. Populated for the SLACK notifications type.
-	DatadogAPIKey       string    `json:"datadogApiKey,omitempty"`       // Datadog API Key. Found in the Datadog dashboard. Populated for the DATADOG notifications type.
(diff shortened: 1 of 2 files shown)


mongodbatlas/alert_configurations.go
```diff
@@ -76,27 +76,29 @@ type MetricThreshold struct {
...
+	TeamID              string    `json:"teamId,omitempty"`              // Unique identifier of a team.
+	TypeName            string    `json:"typeName,omitempty"`            // Type of alert notification.
+	Username            string    `json:"username,omitempty"`            // Name of the Atlas user to which to send notifications. Only a user in the project that owns the alert configuration is allowed here. Populated for the USER notifications type.
+	VictorOpsAPIKey     string    `json:"victorOpsApiKey,omitempty"`     // VictorOps API key. Populated for the VICTOR_OPS notifications type. If the key later becomes invalid, Atlas sends an email to the project owner and eventually removes the key.
+	VictorOpsRoutingKey string    `json:"victorOpsRoutingKey,omitempty"` // VictorOps routing key. Populated for the VICTOR_OPS notifications type. If the key later becomes invalid, Atlas sends an email to the project owner and eventually removes the key.
+	Roles               *[]string `json:"roles,omitempty"`               // The following roles grant privileges within a project.
```
PacoDw: Why did you prefer to use *[]string? I think *[]string is not necessary to use because you have omitempty label,

coderGo93: True that, is not necessary in this case, thank you, paco

themantissa:
LGTM - please ensure at least one DoU approval before merging.  Thank you!!

mongodbatlas/alert_configurations.go
PacoDw: you're welcome :)

PR #71, opened 2020-03-24
Add get processes in a group

mmb:
Co-authored-by: J.P. Poveda <etsuo@users.noreply.github.com>

### DIFF
--- a/mongodbatlas/processes.go
+++ b/mongodbatlas/processes.go
@@ -1,0 +1,74 @@
+package mongodbatlas
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+const processesPath = "groups/%s/processes"
+
+// ProcessesService is for interfacing with the project Processes endpoints of
+// the MongoDB Atlas API.
+// See more: https://docs.atlas.mongodb.com/reference/api/monitoring-and-logs/
+type ProcessesService interface {
+	List(context.Context, string, *ListOptions) ([]*Process, *Response, error)
+}
+
+// ProcessesServiceOp handles communication with the Process related methods
+// of the MongoDB Atlas API.
+type ProcessesServiceOp struct {
+	Client RequestDoer
+}
+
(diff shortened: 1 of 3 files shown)


mmb:
This is a preliminary PR for adding support for the measurements API. Our goal is to use the measurements API to write a prometheus exporter for Atlas.

mongodbatlas/processes.go
```diff
@@ -0,0 +1,65 @@
...
+	Port           int     `json:"port"`
+	ShardName      string  `json:"shardName"`
+	ReplicaSetName string  `json:"replicaSetName"`
+	TypeName       string  `json:"typeName"`
+	Version        string  `json:"version"`
+}
+
+type processesResponse struct {
+	Links      []*Link   `json:"links,omitempty"`
+	Results    []Process `json:"results,omitempty"`
```
gssbzn: [nit] We try to keep the results as pointers, this is also necessary for the `omitempty` in this case

mongodbatlas/processes.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+const processesPath = "groups/%s/processes"
+
+type ProcessesService interface {
```
gssbzn: some small nits with all public methods/structs, we try to keep then documented and following and with links to the api docs (https://docs.atlas.mongodb.com/reference/api/processes-get-all/ in this case) you can follow some other implementations as a reference of how we format these docs

mmb:
Thank you for review! I have added the comments, used pointers and signed the CLA.

Please let us know if there is anything else that can be improved.

mmb:
Thanks all. Great news about the measurements API.

themantissa:
@mmb can you tidy up based on the CI results so we can merge?  Thanks!

themantissa:
@gssbzn thank you so much for that!   @mmb all good now!

PR #79, opened 2020-04-07
CLOUDP-60178: Implement logs API, Atlas

andreaangiolillo:
[CLOUDP-60178](https://jira.mongodb.org/browse/CLOUDP-60178)

### DIFF
--- a/mongodbatlas/logs.go
+++ b/mongodbatlas/logs.go
@@ -13,7 +13,7 @@ const logsPath = "groups/%s/clusters/%s/logs/%s"
 // endpoints of the MongoDB Atlas API.
 // See more: https://docs.atlas.mongodb.com/reference/api/logs/
 type LogsService interface {
-	Get(context.Context, string, string, string, io.Writer, *LogsListOptions) (*Response, error)
+	Get(context.Context, string, string, string, io.Writer, *DateRangetOptions) (*Response, error)
 }
 
 // LogsServiceOp handles communication with the Logs related methods of the
@@ -22,15 +22,16 @@ type LogsServiceOp struct {
 	Client GZipRequestDoer
 }
 
-// LogsListOptions specifies the optional parameters to the LogsService List method.
-type LogsListOptions struct {
+// DateRangetOptions specifies an optional date range query.
+type DateRangetOptions struct {
 	StartDate string `url:"startDate,omitempty"`
 	EndDate   string `url:"endDate,omitempty"`
 }
 
-// Get gets a compressed (.gz) log file that contains a range of log messages for a particular host in an Atlas host.
+// Get gets a compressed (.gz) log file that contains a range of log messages for a particular host.
+// Note: The input parameter out (io.Writer) is not closed by this function.
 // See more: https://docs.atlas.mongodb.com/reference/api/logs/
-func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string, logName string, out io.Writer, opts *LogsListOptions) (*Response, error) {
+func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string, logName string, out io.Writer, opts *DateRangetOptions) (*Response, error) {
 	if groupID == "" {
 		return nil, NewArgError("groupID", "must be set")
 	}
--- a/mongodbatlas/logs.go
+++ b/mongodbatlas/logs.go
@@ -3,7 +3,6 @@ package mongodbatlas
 import (
 	"context"
 	"fmt"
-	"io"
 	"net/http"
 	"os"
 )
@@ -12,7 +11,7 @@ const logsPath = "groups/%s/clusters/%s/logs/%s"
 
 // LogsService is an interface for interfacing with the Logs
 // endpoints of the MongoDB Atlas API.
-// See more: https://docs.atlas.mongodb.com/reference/api/process-measurements/
+// See more: https://docs.atlas.mongodb.com/reference/api/logs/
 type LogsService interface {
 	Get(context.Context, string, string, string, *LogsListOptions) (*Response, error)
 }
@@ -20,7 +19,7 @@ type LogsService interface {
 // LogsServiceOp handles communication with the Logs related methods of the
 // MongoDB Atlas API
 type LogsServiceOp struct {
-	Client RequestDoer
+	Client GZipRequestDoer
 }
 
 // EventListOptions specifies the optional parameters to the Event List methods.
@@ -29,32 +28,9 @@ type LogsListOptions struct {
 	EndDate   string `url:"endDate,omitempty"`
 }
 
-// write as it downloads and not load the whole file into memory.
-func DownloadFile(filepath string, url string) error {
-
-	// Get the data
-	resp, err := http.Get(url)
-	if err != nil {
-		return err
-	}
-	defer resp.Body.Close()
-
-	// Create the file
-	out, err := os.Create(filepath)
-	if err != nil {
-		return err
-	}
-	defer out.Close()
-
-	// Write the body to file
-	_, err = io.Copy(out, resp.Body)
-	return err
-}
-
 // Get gets a compressed (.gz) log file that contains a range of log messages for a particular host in an Atlas cluster.
 // See more: https://docs.atlas.mongodb.com/reference/api/logs/
 func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string, logName string, opts *LogsListOptions) (*Response, error) {
-
 	if groupID == "" {
 		return nil, NewArgError("groupID", "must be set")
 	}
@@ -69,18 +45,25 @@ func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string
 
 	basePath := fmt.Sprintf(logsPath, groupID, hostName, logName)
 
-	//Add query params from listOptions
+	// Add query params
 	path, err := setListOptions(basePath, opts)
 	if err != nil {
 		return nil, err
 	}
 
-	req, err := s.Client.NewGZipRequest(ctx, http.MethodGet, path, nil)
+	req, err := s.Client.NewGZipRequest(ctx, http.MethodGet, path)
 	if err != nil {
 		return nil, err
 	}
 
-	resp, err := s.Client.Do(ctx, req, nil)
+	out, err := os.Create(logName)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := s.Client.Do(ctx, req, out)
+	defer out.Close()
+
 	if err != nil {
 		return resp, err
 	}
--- a/mongodbatlas/logs.go
+++ b/mongodbatlas/logs.go
@@ -28,6 +28,11 @@ type LogsListOptions struct {
 	EndDate   string `url:"endDate,omitempty"`
 }
 
+// Download creates a file with the name specified in logName.
+func (s *LogsServiceOp) Download(logName string) (*os.File, error) {
+	return os.Create(logName)
+}
+
 // Get gets a compressed (.gz) log file that contains a range of log messages for a particular host in an Atlas cluster.
 // See more: https://docs.atlas.mongodb.com/reference/api/logs/
 func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string, logName string, opts *LogsListOptions) (*Response, error) {
@@ -56,7 +61,7 @@ func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string
 		return nil, err
 	}
 
-	out, err := os.Create(logName)
+	out, err := s.Download(logName)
 	if err != nil {
 		return nil, err
 	}
--- a/mongodbatlas/logs.go
+++ b/mongodbatlas/logs.go
@@ -75,7 +75,7 @@ func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string
 		return nil, err
 	}
 
-	req, err := s.Client.NewRequest(ctx, http.MethodGet, path, nil)
+	req, err := s.Client.NewGZipRequest(ctx, http.MethodGet, path, nil)
 	if err != nil {
 		return nil, err
 	}
--- a/mongodbatlas/logs.go
+++ b/mongodbatlas/logs.go
@@ -51,7 +51,7 @@ func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string
 		return nil, err
 	}
 
-	req, err := s.Client.NewGZipRequest(ctx, http.MethodGet, path, nil)
+	req, err := s.Client.NewGZipRequest(ctx, http.MethodGet, path)
 	if err != nil {
 		return nil, err
 	}
--- a/mongodbatlas/logs.go
+++ b/mongodbatlas/logs.go
@@ -1,0 +1,89 @@
+package mongodbatlas
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+)
+
+const logsPath = "groups/%s/clusters/%s/logs/%s"
+
+// LogsService is an interface for interfacing with the Logs
+// endpoints of the MongoDB Atlas API.
+// See more: https://docs.atlas.mongodb.com/reference/api/process-measurements/
+type LogsService interface {
+	Get(context.Context, string, string, string, *LogsListOptions) (*Response, error)
+}
+
+// LogsServiceOp handles communication with the Logs related methods of the
+// MongoDB Atlas API
+type LogsServiceOp struct {
+	Client RequestDoer
+}
+
+// EventListOptions specifies the optional parameters to the Event List methods.
+type LogsListOptions struct {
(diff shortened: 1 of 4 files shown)


mongodbatlas/logs.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := s.Client.NewGZipRequest(ctx, http.MethodGet, path)
+	if err != nil {
+		return nil, err
+	}
+
+	out, err := os.Create(logName)
```
gssbzn, line 59: Right now this method is creating the output file but I think we should pass it the file, an `io.Writer` probably or accept the name of the output file I don't know which one is better

mongodbatlas/mongodbatlas.go
```diff
@@ -233,29 +250,56 @@ func SetUserAgent(ua string) ClientOpt {
...
-	rel, err := url.Parse(urlStr)
+	u, err := c.BaseURL.Parse(urlStr)
 	if err != nil {
 		return nil, err
 	}
 
-	u := c.BaseURL.ResolveReference(rel)
-
-	buf := new(bytes.Buffer)
+	var buf io.ReadWriter
```
gssbzn, line 258: we don't need buf unless there's a body

mongodbatlas/mongodbatlas.go
```diff
@@ -233,29 +250,56 @@ func SetUserAgent(ua string) ClientOpt {
...
 	}
 
-	req, err := http.NewRequest(method, u.String(), buf)
+	req, err := http.NewRequestWithContext(ctx, method, u.String(), buf)
 	if err != nil {
 		return nil, err
 	}
 
-	req.Header.Add("Content-Type", mediaType)
+	if body != nil {
```
gssbzn, line 274: only set the Content-Type if there's a body

mongodbatlas/mongodbatlas.go
```diff
@@ -270,6 +314,14 @@ func (c *Client) OnRequestCompleted(rc RequestCompletionCallback) {
 func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Response, error) {
 	resp, err := DoRequestWithClient(ctx, c.client, req)
```
gssbzn, line 315: I would like to get rid of `DoRequestWithClient` since it doe snot offer any advantage but I don't know if someone is using it

mongodbatlas/mongodbatlas.go
```diff
@@ -296,13 +348,15 @@ func (c *Client) Do(ctx context.Context, req *http.Request, v interface{}) (*Res
 				return nil, err
 			}
 		} else {
-			err = json.NewDecoder(resp.Body).Decode(v)
-			if err != nil {
-				return nil, err
+			decErr := json.NewDecoder(resp.Body).Decode(v)
+			if decErr == io.EOF {
```
gssbzn, line 352: this is a small bug fix I did for om, bringing it here

mongodbatlas/logs.go
andreaangiolillo: why do you think so? I do not see any advantages in changing it. It seems really clean in this way.

mongodbatlas/mongodbatlas.go
```diff
@@ -233,29 +250,71 @@ func SetUserAgent(ua string) ClientOpt {
...
 			return nil, err
 		}
 	}
 
+	return buf, nil
+}
+
+// NewGZipRequest creates an API request that accepts gzip. A relative URL can be provided in urlStr, which will be resolved to the
+// BaseURL of the Client. Relative URLS should always be specified without a preceding slash.
+func (c *Client) NewGZipRequest(ctx context.Context, method, urlStr string, body interface{}) (*http.Request, error) {
```
gssbzn, line 296: [nit] the reason I did not have a body here is that I expect all `NewGZipRequest` to be GETs, gets don't accept a body so no need for it at all

mongodbatlas/mongodbatlas.go
```diff
@@ -233,29 +250,71 @@ func SetUserAgent(ua string) ClientOpt {
...
+	rel, err := url.Parse(urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	u := c.BaseURL.ResolveReference(rel)
+
+	buf, err := c.ReadBody(body)
```
gssbzn, line 304: with that said of not requiring a body there's no need for this

mongodbatlas/mongodbatlas.go
```diff
@@ -233,29 +250,71 @@ func SetUserAgent(ua string) ClientOpt {
...
+	}
+	req.Header.Add("Accept", mediaType)
+	if c.UserAgent != "" {
+		req.Header.Set("User-Agent", c.UserAgent)
+	}
+	return req, nil
+}
+
+// ReadBody returns an ReadWriter object containing the body for the http request
+func (c *Client) ReadBody(body interface{}) (io.ReadWriter, error) {
```
gssbzn, line 279: [nit] couple of things, this is not "reading a body as much as it is encoding it or serialising it, which would be a more correct action, this is the body of the client request, not from the server response
I guess you created the method to reuse it in both NewRequest and NewGZipRequest, like I mention before `NewGZipRequest` should not need this, so you could delete this method
If you decide to keep it please consider to make it private and not public along with the name change

mongodbatlas/logs.go
gssbzn: the main reason to change it is that users don't have control over the name file which may or may not be desirable, that's why at least passing the output file name should be enough
on the other hand the problem with using `os.Create(logName)` directly is that we depend on a particular implementation of `io.Writer` which is kind of bad as not all the users of this client may actually have access to a file system but may still have a way to write files
I think an ideal world we could have this `Get` method with an `io.Writer` param, and then we can create a `Download` method which does this assuming `os.Create(logName)`

mongodbatlas/mongodbatlas_test.go
```diff
@@ -153,6 +171,68 @@ func TestNewRequest_withCustomUserAgent(t *testing.T) {
...
+	}
+}
+
+func TestNewGZipRequest_withUserData(t *testing.T) {
+	c := NewClient(nil)
+
+	requestPath := "foo"
+
+	inURL, outURL := requestPath, defaultBaseURL+requestPath
+	inBody, outBody := &testRequestBody{TestName: "l", TestData: "u"},
```
gssbzn, line 191: this tests is a lie as you don't expect a json response (nor a json payload as already discussed), let's just make sure that the headers are correct

mongodbatlas/logs.go
andreaangiolillo: Do you want me to make this change (`io.Writer` as param)? I am not sure how to implement the download method though

gssbzn: yes, let's create the two method in this PR as getting this merge takes some time so we don't have to revisit this in case we need it.
on how to implement we should split the current method in two, get should not do the os.create but directly pass the parma to the Do method and thus makeing this method a bit simpler
Then the download method should take the logName and create a file, just like we are doing here now, and then with that new file call the simpler Get method, does that make sense?

mongodbatlas/mongodbatlas_test.go
```diff
@@ -153,6 +171,39 @@ func TestNewRequest_withCustomUserAgent(t *testing.T) {
 	}
 }
 
+func TestNewGZipRequest_emptyBody(t *testing.T) {
```
gssbzn, line 174: can we also add a test to check teh default behaviour, something like
```go
func TestNewGZipRequest(t *testing.T) {
	c := NewClient(nil)

	requestPath := "foo"

	inURL, outURL := requestPath, defaultBaseURL+requestPath
	req, _ := c.NewGZipRequest(ctx, http.MethodGet, inURL)

	// test relative URL was expanded
	if req.URL.String() != outURL {
		t.Errorf("NewGZipRequest(%v) URL = %v, expected %v", inURL, req.URL, outURL)
	}

	// test accept content type is correct
	accept := req.Header.Get("Accept")
	if gzipMediaType != accept {
		t.Errorf("NewGZipRequest() Accept = %v, expected %v", accept, gzipMediaType)
	}

	// test default user-agent is attached to the request
	userAgent := req.Header.Get("User-Agent")
	if c.UserAgent != userAgent {
		t.Errorf("NewGZipRequest() User-Agent = %v, expected %v", userAgent, c.UserAgent)
	}
}
```

mongodbatlas/logs.go
andreaangiolillo: Just to be sure that we are on the same page, I am going to add the writer to the `Get` but I will not remove the logName( I may change its name) because according to the  documentation it allows to retrieve different type of log file:
- mongodb.gz
- mongos.gz
- mongodb-audit-log.gz
- mongos-audit-log.gz

mongodbatlas/logs.go
```diff
@@ -0,0 +1,77 @@
...
+	Get(context.Context, string, string, string, *LogsListOptions) (*Response, error)
+}
+
+// LogsServiceOp handles communication with the Logs related methods of the
+// MongoDB Atlas API
+type LogsServiceOp struct {
+	Client GZipRequestDoer
+}
+
+// EventListOptions specifies the optional parameters to the Event List methods.
```
gssbzn, line 25:
```suggestion
// LogsListOptions specifies the optional parameters to the LogsService List method.
```

mongodbatlas/logs.go
```diff
@@ -0,0 +1,77 @@
...
+	StartDate string `url:"startDate,omitempty"`
+	EndDate   string `url:"endDate,omitempty"`
+}
+
+// Download creates a file with the name specified in logName.
+func (s *LogsServiceOp) Download(logName string) (*os.File, error) {
+	return os.Create(logName)
+}
+
+// Get gets a compressed (.gz) log file that contains a range of log messages for a particular host in an Atlas cluster.
```
gssbzn, line 36:
```suggestion
// Get gets a compressed (.gz) log file that contains a range of log messages for a particular host in an Atlas host.
```

mongodbatlas/logs.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := s.Client.NewGZipRequest(ctx, http.MethodGet, path)
+	if err != nil {
+		return nil, err
+	}
+
+	out, err := s.Download(logName)
```
gssbzn, line 64: I thought we agreed we were going to make Download call Get, not the other way around

mongodbatlas/logs_test.go
```diff
@@ -0,0 +1,25 @@
+package mongodbatlas
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+)
+
+func TestLogs_Get(t *testing.T) {
```
gssbzn, line 9: we should have a test for `Download`

mongodbatlas/logs.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// Download creates a file with the name specified in logName.
+func (s *LogsServiceOp) Download(logName string) (*os.File, error) {
+	return os.Create(logName)
+}
+
+// Get gets a compressed (.gz) log file that contains a range of log messages for a particular host in an Atlas cluster.
+// See more: https://docs.atlas.mongodb.com/reference/api/logs/
+func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string, logName string, opts *LogsListOptions) (*Response, error) {
```
gssbzn, line 38: I thought we agree this would be something like
```suggestion
func (s *LogsServiceOp) Get(ctx context.Context, groupID string, hostName string, logName string, opts *LogsListOptions, out io.Writer) (*Response, error) {
```
And download would make use to that `out` param passing the create file system

andreaangiolillo:
I still have to implement the test for `Download` but, in the meantime, I would like to know if this implementation of the `Get` and `Download` is what you want 😸 .

mongodbatlas/logs.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// LogsServiceOp handles communication with the Logs related methods of the
+// MongoDB Atlas API
+type LogsServiceOp struct {
+	Client GZipRequestDoer
+}
+
+// LogsListOptions specifies the optional parameters to the LogsService List method.
+type LogsListOptions struct {
```
gssbzn, line 26:
```suggestion
// DateRangetOptions specifies an optional date range query.
type DateRangetOptions struct {
```

mongodbatlas/logs.go
```diff
@@ -0,0 +1,66 @@
...
+	StartDate string `url:"startDate,omitempty"`
+	EndDate   string `url:"endDate,omitempty"`
+}
+
+// Get gets a compressed (.gz) log file that contains a range of log messages for a particular host in an Atlas host.
```
gssbzn, line 31:
```suggestion
// Get gets a compressed (.gz) log file that contains a range of log messages for a particular host.
```

gssbzn: Can we also add a note that `out` won't be closed and that users should do it?

mongodbatlas/logs_test.go
```diff
@@ -0,0 +1,25 @@
...
+	setup()
+	defer teardown()
+
+	groupID := "1"
+	cluster := "test-username"
+	log := "log"
+
+	mux.HandleFunc(fmt.Sprintf("/groups/%s/clusters/%s/logs/%s", groupID, cluster, log), func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, http.MethodGet)
```
gssbzn, line 18: [q] can you make this return a random text?
then you can do something like
```
buf := new(bytes.Buffer)
client.Logs.Get(ctx, groupID, cluster, log, buf, nil)
if buf.String() != "random string" {
 t.Error....
}
```
(advantages of depending on an interface ;))

mongodbatlas/mongodbatlas.go
```diff
@@ -239,29 +256,66 @@ func SetUserAgent(ua string) ClientOpt {
...
+	if c.UserAgent != "" {
+		req.Header.Set("User-Agent", c.UserAgent)
+	}
+	return req, nil
+}
+
+// serializeRequestBody returns an ReadWriter object containing the body of the http request
+func (c *Client) serializeRequestBody(body interface{}) (io.ReadWriter, error) {
+	var buf io.ReadWriter
+	if body != nil {
```
gssbzn, line 287: this if should be done before calling this method and if you do that then it makes sense to call this method `newEncodedBody`

mongodbatlas/mongodbatlas.go
```diff
@@ -239,29 +256,66 @@ func SetUserAgent(ua string) ClientOpt {
...
+	}
+	req.Header.Add("Accept", mediaType)
+	if c.UserAgent != "" {
+		req.Header.Set("User-Agent", c.UserAgent)
+	}
+	return req, nil
+}
+
+// serializeRequestBody returns an ReadWriter object containing the body of the http request
+func (c *Client) serializeRequestBody(body interface{}) (io.ReadWriter, error) {
```
gssbzn, line 285: [q] have you tried to return a `io.Reader` instead? it looks like `http.NewRequest(method, u.String(), buf)` only needs a reader

mongodbatlas/mongodbatlas.go
```diff
@@ -281,17 +284,14 @@ func (c *Client) NewRequest(ctx context.Context, method, urlStr string, body int
...
-		if err != nil {
-			return nil, err
-		}
+// newEncodedBody returns an ReadWriter object containing the body of the http request
+func (c *Client) newEncodedBody(body interface{}) (io.Reader, error) {
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	err := enc.Encode(body)
+	if err != nil {
```
gssbzn, line 293: no need for the if, just always do
```go
 return buf, err
```

mongodbatlas/mongodbatlas.go
```diff
@@ -260,8 +260,11 @@ func (c *Client) NewRequest(ctx context.Context, method, urlStr string, body int
 	if err != nil {
 		return nil, err
 	}
+	var buf io.Reader
+	if body != nil {
+		buf, err = c.newEncodedBody(body)
```
gssbzn, line 265: you should move the check for error here
```suggestion
		if buf, err = c.newEncodedBody(body); err !=nil{
 			return nil, err
		}
```

gssbzn:
@themantissa this is ready for review, it includes some changes on how requests are created/executed to let us handle gzip request so you may want to give that some extra attention

themantissa:
@PacoDw  and @marinsalinas as the two most familiar with this client and all the downstream consumers ( Terraform provider, Vault and CloudFormation) can you in particular review the changes in  mongodbatlas.go, and mongodbatlas_test.go, especially around the new request interface, etc?   This may take a bit longer so if we need to use retainer hours that is fine.

gssbzn:
@marinsalinas yes, we don't have POST APIs that return gzip content

PR #83, opened 2020-04-21
Validate base url trailing slash

gssbzn:
## Proposed changes

This is not so much a problem for Atlas but for Ops Manager where you set a custom url we are having users face this problem

This change is two fold, we validate the request to be correct and we ensure in tests that devs use a relative path and not an absolute one

## Checklist

### DIFF
--- a/mongodbatlas/alert_configurations_test.go
+++ b/mongodbatlas/alert_configurations_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 func TestAlertConfiguration_Create(t *testing.T) {
-	setup()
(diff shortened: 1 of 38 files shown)


gssbzn:
@themantissa this is not urgent, just a nice to have that we added in the OM client and made sense to add here as well as we use some of the atlas services

themantissa:
@PacoDw @marinsalinas if you have a second to just give this a look and give it a thumbs up that would be great.  Let me know if I should create a Trello card to track it, should just be a quick one.

PR #84, opened 2020-04-23
Fix cluster struct to allow pasue/start a cluster

gssbzn:
## Proposed changes

**Breaking change**

Currently if you want to pause or restart a cluster you can only send the `Paused` property and nothing else or you'll get errors like "Can't pause and update a cluster at the same time"

Having some internal structs as non-pointers were making go serialize them so we could never do this kind of operation

Sadly this may be a breaking change as we are changing the type of the struct and those depending on it may need to update after this change

## Checklist

### DIFF
--- a/mongodbatlas/clusters.go
+++ b/mongodbatlas/clusters.go
@@ -81,9 +81,9 @@ type ConnectionStrings struct {
 
 // Cluster represents MongoDB cluster.
 type Cluster struct {
-	AutoScaling              AutoScaling              `json:"autoScaling,omitempty"`
+	AutoScaling              *AutoScaling             `json:"autoScaling,omitempty"`
 	BackupEnabled            *bool                    `json:"backupEnabled,omitempty"`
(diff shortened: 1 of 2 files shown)


gssbzn:
@themantissa tagging you in advance as this may be a breaking change but sadly is a bug that we may need to fix sooner than later (this came from a customer that wants to pause/restart clusters with the cli)

themantissa:
Thank you for the alert @gssbzn - makes sense for the CLI.  We haven't yet supported in Terraform since Atlas auto unpauses, which doesn't quite mesh w/ the Terraform philosophy.

@PacoDw and @marinsalinas (and @thetonymaster) I assume this will impact both Terraform and CloudFormation so we'll need to schedule the work to coordinate with the next release of the go-client?

PacoDw:
yes @themantissa, it will, but it won't be a breaking change for terraform

themantissa:
@PacoDw can you also just submit a review?

gssbzn:
Thank yall for the review

PR #85, opened 2020-04-23
Make the measurements param an array

gssbzn:
## Proposed changes
the `m` param is actually an array so fixing this

## Checklist


gssbzn:
@themantissa this is only used by cli so I'll merge this away

themantissa:
Makes total sense but thank you for the tag/info @gssbzn !

PR #87, opened 2020-05-01
Use the new vanity url

gssbzn:
## Proposed changes

We have set up these urls to resolve to this package, this is up for consideration so let me know what you think


themantissa:
Approving as fine for now.

PR #91, opened 2020-05-08
Three new fields for Snapshot restore jobs options

coderGo93:
Added new three fields for Snapshot restore jobs called `oplogTs`, `oplogInc` and `pointInTimeUTCSeconds` and its test

### DIFF
--- a/mongodbatlas/cloud_provider_snapshot_restore_jobs.go
+++ b/mongodbatlas/cloud_provider_snapshot_restore_jobs.go
@@ -30,19 +30,21 @@ var _ CloudProviderSnapshotRestoreJobsService = &CloudProviderSnapshotRestoreJob
 
 // CloudProviderSnapshotRestoreJob represents the structure of a cloudProviderSnapshotRestoreJob.
 type CloudProviderSnapshotRestoreJob struct {
-	ID                string   `json:"id,omitempty"`                // The unique identifier of the restore job.
-	SnapshotID        string   `json:"snapshotId,omitempty"`        // Unique identifier of the snapshot to restore.
-	DeliveryType      string   `json:"deliveryType,omitempty"`      // Type of restore job to create. Possible values are: automated or download
-	DeliveryURL       []string `json:"deliveryUrl,omitempty"`       // One or more URLs for the compressed snapshot files for manual download. Only visible if deliveryType is download.
-	TargetClusterName string   `json:"targetClusterName,omitempty"` // Name of the target Atlas cluster to which the restore job restores the snapshot. Only required if deliveryType is automated.
-	TargetGroupID     string   `json:"targetGroupId,omitempty"`     // Unique ID of the target Atlas project for the specified targetClusterName. Only required if deliveryType is automated.
-	Cancelled         bool     `json:"cancelled,omitempty"`         // Indicates whether the restore job was canceled.
-	CreatedAt         string   `json:"createdAt,omitempty"`         // UTC ISO 8601 formatted point in time when Atlas created the restore job.
-	Expired           bool     `json:"expired,omitempty"`           // Indicates whether the restore job expired.
-	ExpiresAt         string   `json:"expiresAt,omitempty"`         // UTC ISO 8601 formatted point in time when the restore job expires.
-	FinishedAt        string   `json:"finishedAt,omitempty"`        // UTC ISO 8601 formatted point in time when the restore job completed.
-	Links             []*Link  `json:"links,omitempty"`             // One or more links to sub-resources and/or related resources. The relations between URLs are explained in the Web Linking Specification.
-	Timestamp         string   `json:"timestamp,omitempty"`         // Timestamp in ISO 8601 date and time format in UTC when the snapshot associated to snapshotId was taken.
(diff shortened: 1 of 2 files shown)


themantissa:
@gssbzn added you for a review as an FYI and check in case it's something you can use for mCLI.

mongodbatlas/cloud_provider_snapshot_restore_jobs.go
```diff
@@ -30,19 +30,21 @@ var _ CloudProviderSnapshotRestoreJobsService = &CloudProviderSnapshotRestoreJob
...
+	CreatedAt             string   `json:"createdAt,omitempty"`             // UTC ISO 8601 formatted point in time when Atlas created the restore job.
+	Expired               bool     `json:"expired,omitempty"`               // Indicates whether the restore job expired.
+	ExpiresAt             string   `json:"expiresAt,omitempty"`             // UTC ISO 8601 formatted point in time when the restore job expires.
+	FinishedAt            string   `json:"finishedAt,omitempty"`            // UTC ISO 8601 formatted point in time when the restore job completed.
+	Links                 []*Link  `json:"links,omitempty"`                 // One or more links to sub-resources and/or related resources. The relations between URLs are explained in the Web Linking Specification.
+	Timestamp             string   `json:"timestamp,omitempty"`             // Timestamp in ISO 8601 date and time format in UTC when the snapshot associated to snapshotId was taken.
+	OplogTs               string   `json:"oplogTs,omitempty"`               // If your cluster has Point-in-Time restores enabled, you can specify a timestamp from the oplog from which to restore a backup.
```
gssbzn, line 46: had to go and compare our docs to our code, they both should be int64, and for your test the sample should be something like `1536610288`

gssbzn: We may need to file a ticket with the docs team to fix this cc @themantissa

gssbzn, line 46: Another thing here, if you provide `oplogTs` you **must** also provide an `oplogInc` (int)
this also looks to be missing from the docs

themantissa: @gssbzn huge thank you for catching this!  I'll get a docs ticket in.

coderGo93: I understand, gustavo, I'll change it and thank you very much :+1:

coderGo93: I see, makes sense, gustavo, thank you very much, I'll change it

mongodbatlas/cloud_provider_snapshot_restore_jobs.go
```diff
@@ -30,19 +30,21 @@ var _ CloudProviderSnapshotRestoreJobsService = &CloudProviderSnapshotRestoreJob
...
-	Expired           bool     `json:"expired,omitempty"`           // Indicates whether the restore job expired.
-	ExpiresAt         string   `json:"expiresAt,omitempty"`         // UTC ISO 8601 formatted point in time when the restore job expires.
-	FinishedAt        string   `json:"finishedAt,omitempty"`        // UTC ISO 8601 formatted point in time when the restore job completed.
-	Links             []*Link  `json:"links,omitempty"`             // One or more links to sub-resources and/or related resources. The relations between URLs are explained in the Web Linking Specification.
-	Timestamp         string   `json:"timestamp,omitempty"`         // Timestamp in ISO 8601 date and time format in UTC when the snapshot associated to snapshotId was taken.
+	ID                    string   `json:"id,omitempty"`                    // The unique identifier of the restore job.
+	SnapshotID            string   `json:"snapshotId,omitempty"`            // Unique identifier of the snapshot to restore.
+	DeliveryType          string   `json:"deliveryType,omitempty"`          // Type of restore job to create. Possible values are: automated or download
```
themantissa, line 35: May also want to update the comment - there are 3 types now, this is the newest: pointInTime

mongodbatlas/cloud_provider_snapshot_restore_jobs_test.go
```diff
@@ -161,8 +163,10 @@ func TestCloudProviderSnapshotRestoreJobs_List(t *testing.T) {
 						Rel:  "http://mms.mongodb.com/group",
 					},
 				},
-				SnapshotID: "5b6211ff87d9d663c59d3feg",
-				Timestamp:  "2018-08-01T20:02:07Z",
+				SnapshotID:            "5b6211ff87d9d663c59d3feg",
+				Timestamp:             "2018-08-01T20:02:07Z",
+				OplogTs:               "2018-08-01T20:02:07Z",
+				PointInTimeUTCSeconds: 1583753751,
```
themantissa, line 169: Also fyi for later be sure when supporting in Terraform you note that deliveryType for doing a point in time restore job (PIT) would be pointInTime also note in the docs that you must specify either pointInTimeUTCSeconds or oplogTs, but not both.

mongodbatlas/cloud_provider_snapshot_restore_jobs.go
coderGo93: You're right, I'll change it, thank you melissa

PR #92, opened 2020-05-18
Cluster autoscaling compute

coderGo93:
Added fields for autoscaling
* Compute

Added a new sruct Compute and its fields
* Enabled
* ScaleDownEnabled
* MinInstanceSize
* MaxInstanceSize

### DIFF
--- a/mongodbatlas/clusters_test.go
+++ b/mongodbatlas/clusters_test.go
@@ -20,7 +20,11 @@ func TestClusters_ListClusters(t *testing.T) {
 			"results": [
 				{
 					"autoScaling": {
-						"diskGBEnabled": true
+						"diskGBEnabled": true,
+						"compute": {
+						  "enabled": true,
+						  "scaleDownEnabled": true
+						}
 					},
 					"backupEnabled": true,
 					"biConnector": {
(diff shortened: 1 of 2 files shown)


mongodbatlas/clusters_test.go
```diff
@@ -55,7 +59,13 @@ func TestClusters_ListClusters(t *testing.T) {
 						"diskIOPS": 1320,
 						"encryptEBSVolume": false,
 						"instanceSizeName": "M40",
-						"regionName": "US_WEST_2"
+						"regionName": "US_WEST_2",
+						"autoScaling": {
+							"compute": {
+							  "maxInstanceSize": "M40",
+							  "minInstanceSize": "M10"
```
themantissa, line 66: It doesn't really make sense to have the max size match the instanceSizeName, if we want to make it a bit more a real world example max is usually a cluster size or two (on average) more than the instanceSizeName a user starts with and min one or two below - M60 would make more sense in this case.

coderGo93: You're right, I'll change it to M60 in max since is the same and it should be higher than instanceSizeName like you said, thank you very much melissa

PR #108, opened 2020-07-08
CLOUDP-66607: Atlas Data Lakes

robcarlan-mlab:
## Description

Adds data lake support.

Link to any related issue(s): 

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/data_lakes.go
+++ b/mongodbatlas/data_lakes.go
@@ -45,7 +45,7 @@ type DataLakeStore struct {
 	Bucket      string `json:"bucket,omitempty"`
 	Prefix      string `json:"prefix,omitempty"`
 	Delimiter   string `json:"delimiter,omitempty"`
-	IncludeTags bool   `json:"includeTags,omitempty"`
+	IncludeTags *bool  `json:"includeTags,omitempty"`
 }
 
 type DataLakeDataSource struct {
--- a/mongodbatlas/data_lakes.go
+++ b/mongodbatlas/data_lakes.go
@@ -135,7 +135,7 @@ func (s *DataLakeServiceOp) List(ctx context.Context, groupID string) ([]DataLak
 // Get gets the data laked associated with a specific name.
 //
 // See more: https://docs.mongodb.com/datalake/reference/api/dataLakes-get-one-tenant/
-func (s *DataLakeServiceOp) Get(ctx context.Context, groupID string, name string) (*DataLake, *Response, error) {
+func (s *DataLakeServiceOp) Get(ctx context.Context, groupID, name string) (*DataLake, *Response, error) {
 	if groupID == "" {
 		return nil, nil, NewArgError("groupID", "must be set")
 	}
@@ -189,7 +189,7 @@ func (s *DataLakeServiceOp) Create(ctx context.Context, groupID string, createRe
 // Update updates an existing Data Lake.
 //
 // See more: https://docs.mongodb.com/datalake/reference/api/dataLakes-update-one-tenant/
-func (s *DataLakeServiceOp) Update(ctx context.Context, groupID string, name string, updateRequest *DataLakeUpdateRequest) (*DataLake, *Response, error) {
+func (s *DataLakeServiceOp) Update(ctx context.Context, groupID, name string, updateRequest *DataLakeUpdateRequest) (*DataLake, *Response, error) {
 	if groupID == "" {
 		return nil, nil, NewArgError("groupID", "must be set")
 	}
@@ -219,7 +219,7 @@ func (s *DataLakeServiceOp) Update(ctx context.Context, groupID string, name str
 // Delete deletes the Data Lake with a given name.
 //
 // See more: https://docs.mongodb.com/datalake/reference/api/dataLakes-delete-one-tenant/
-func (s *DataLakeServiceOp) Delete(ctx context.Context, groupID string, name string) (*Response, error) {
+func (s *DataLakeServiceOp) Delete(ctx context.Context, groupID, name string) (*Response, error) {
 	if groupID == "" {
 		return nil, NewArgError("groupId", "must be set")
 	}
--- a/mongodbatlas/data_lakes.go
+++ b/mongodbatlas/data_lakes.go
@@ -11,9 +11,10 @@ const (
 )
 
 // DataLakeService is an interface for interfacing with the Data Lake endpoints of the MongoDB Atlas API.
+//
 // See more: https://docs.mongodb.com/datalake/reference/api/datalakes-api
 type DataLakeService interface {
-	List(context.Context, *DataLakeReqPathParameters, *ListOptions) ([]DataLake, *Response, error)
+	List(context.Context, *DataLakeReqPathParameters) ([]DataLake, *Response, error)
 	Get(context.Context, *DataLakeReqPathParameters) (*DataLake, *Response, error)
 	Create(context.Context, *DataLakeReqPathParameters, *DataLakeCreateRequest) (*DataLake, *Response, error)
 	Update(context.Context, *DataLakeReqPathParameters, *DataLakeUpdateRequest) (*DataLake, *Response, error)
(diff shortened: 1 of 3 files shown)


mongodbatlas/data_lakes.go
```diff
@@ -0,0 +1,236 @@
...
+		return nil, nil, err
+	}
+
+	req, err := s.Client.NewRequest(ctx, http.MethodGet, path, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	root := new([]DataLake)
+	resp, err := s.Client.Do(ctx, req, root)
```
gssbzn, line 127: you may want to do 
```suggestion
	var root []DataLake
	resp, err := s.Client.Do(ctx, req, &root)
```
instead, it's a more common idiom for empty slices and also saves you from the `return *root` to just `return root`

mongodbatlas/data_lakes.go
```diff
@@ -0,0 +1,236 @@
...
+	"fmt"
+	"net/http"
+)
+
+const (
+	dataLakesBasePath = "groups"
+)
+
+// DataLakeService is an interface for interfacing with the Data Lake endpoints of the MongoDB Atlas API.
+// See more: https://docs.mongodb.com/datalake/reference/api/datalakes-api
```
gssbzn, line 14:
```suggestion
// DataLakeService provides access to the data lake related functions in the Atlas API.
//
// See more: https://docs.mongodb.com/datalake/reference/api/datalakes-api
```

please note the extra empty line. When rendering the docs go will ignore new lines (kind of like markdown) so this would be rendered as 
```
DataLakeService is an interface for interfacing with the Data Lake endpoints of the MongoDB Atlas API. See more: https://docs.mongodb.com/datalake/reference/api/datalakes-api
```

I want to slowly fix the existing docs but it may take a while before we get there

mongodbatlas/data_lakes.go
```diff
@@ -0,0 +1,236 @@
...
+	Create(context.Context, *DataLakeReqPathParameters, *DataLakeCreateRequest) (*DataLake, *Response, error)
+	Update(context.Context, *DataLakeReqPathParameters, *DataLakeUpdateRequest) (*DataLake, *Response, error)
+	Delete(context.Context, *DataLakeReqPathParameters) (*Response, error)
+}
+
+// DataLakeServiceOp handles communication with the DataLakeService related methods of the
+// MongoDB Atlas API
+type DataLakeServiceOp service
+
+type AwsCloudProviderConfig struct {
```
gssbzn, line 27: [nit] For public structs we try to keep them documented (some older versions of go vet used to complain about it)
```suggestion
// AwsCloudProviderConfig is the ...
type AwsCloudProviderConfig struct {
```
Same for all other exported structs

mongodbatlas/data_lakes.go
```diff
@@ -0,0 +1,236 @@
...
+)
+
+const (
+	dataLakesBasePath = "groups"
+)
+
+// DataLakeService is an interface for interfacing with the Data Lake endpoints of the MongoDB Atlas API.
+// See more: https://docs.mongodb.com/datalake/reference/api/datalakes-api
+type DataLakeService interface {
+	List(context.Context, *DataLakeReqPathParameters, *ListOptions) ([]DataLake, *Response, error)
```
gssbzn, line 16: reading the docs it doesn't look like we support `ListOptions` for this endpoint

mongodbatlas/data_lakes.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// Create creates a new Data Lake.
+// See more: https://docs.mongodb.com/datalake/reference/api/dataLakes-create-one-tenant/
+func (s *DataLakeServiceOp) Create(ctx context.Context, requestParameters *DataLakeReqPathParameters, createRequest *DataLakeCreateRequest) (*DataLake, *Response, error) {
+	if requestParameters.GroupID == "" {
+		return nil, nil, NewArgError("groupId", "must be set")
+	}
+	if createRequest == nil {
+		return nil, nil, NewArgError("name", "must be set")
```
gssbzn, line 168:
```suggestion
		return nil, nil, NewArgError("createRequest", "must be set")
```

mongodbatlas/data_lakes.go
```diff
@@ -0,0 +1,236 @@
...
+	Hostnames           []string            `json:"hostnames,omitempty"`           // List of hostnames for the data lake.
+	Name                string              `json:"name,omitempty"`                // Name of the data lake.
+	State               string              `json:"state,omitempty"`               // Current state of the data lake.
+	Storage             Storage             `json:"storage,omitempty"`             // Configuration for each data store and its mapping to MongoDB collections / databases.
+}
+
+// DataLakeReqPathParameters represents all the possible parameters to make the request
+type DataLakeReqPathParameters struct {
+	GroupID string `json:"groupId,omitempty"` // The unique identifier of the project for the Atlas cluster.
+	Name    string `json:"name,omitempty"`    // The name of the Data Lake.
```
gssbzn, line 93: [q] given this is used to construct the url do we need the json tags?
[aside] I'm not a fan of putting the URL params as a struct and I'd rather just have explicit params like `func Get(groupId, name string)` but this is not a blocker since I know we have a mix of both approaches around

mongodbatlas/data_lakes.go
```diff
@@ -0,0 +1,236 @@
...
+	DataProcessRegion   DataProcessRegion   `json:"dataProcessRegion,omitempty"`
+}
+
+// DataLakeReqPathParameters represents the required fields to create a new data lake
+type DataLakeCreateRequest struct {
+	Name string `json:"name,omitempty"`
+}
+
+// List gets all data lakes for the specified group.
+// See more: https://docs.mongodb.com/datalake/reference/api/dataLakes-get-all-tenants
```
gssbzn, line 108: Similar to other comments above lets add an extra empty line here

```suggestion
// List gets all data lakes for the specified group.
//
// See more: https://docs.mongodb.com/datalake/reference/api/dataLakes-get-all-tenants
```
for this and all other methods with a "See more" link

mongodbatlas/data_lakes_test.go
```diff
@@ -0,0 +1,348 @@
...
+			Stores:    []DataLakeStore{},
+		},
+	}
+
+	if diff := deep.Equal(cloudProviderSnapshot, &expected); diff != nil {
+		t.Error(diff)
+	}
+}
+
+func TestDataLake_Update(t *testing.T) {
```
gssbzn, line 220: I think we are missing a test for create

mongodbatlas/mongodbatlas.go
```diff
@@ -90,6 +90,7 @@ type Client struct {
 	ProcessDatabaseMeasurements         ProcessDatabaseMeasurementsService
 	Indexes                             IndexesService
 	Logs                                LogsService
+	DataLakes                           DataLakeService
```
gssbzn, line 93: we're going to have conflicts later 😅

mongodbatlas/data_lakes.go
robcarlan-mlab: Yeah we can do without these -- for the struct yeah I wasn't sure if we have a preferred approach so I can go with the other approach if thats slightly preferred

gssbzn: @robcarlan-mlab up to you if you want to refactor, also consider that some of the functions that use this struct only check for the group and not the name but like I mentioned we have a bit of a mix on both so I won't block on this if you prefer to keep it

robcarlan-mlab: I just refactored it :)

mongodbatlas/data_lakes_test.go
```diff
@@ -0,0 +1,412 @@
+package mongodbatlas
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/openlyinc/pointy"
```
gssbzn, line 6: this file is not go imported, I think our `make fmt` may be missing that, travis is failing on this

themantissa:
Added myself to the review list :)

PR #115, opened 2020-07-21
Add a hook to ensure the client's version is updated before pushing a tag/release

MihaiBojin:
## Description

- Versioned the user agent string

Link to any related issue(s): 

## Type of change:

## Required Checklist:


mongodbatlas/mongodbatlas_test.go
```diff
@@ -93,8 +93,8 @@ func testClientDefaultBaseURL(t *testing.T, c *Client) {
 }
 
 func testClientDefaultUserAgent(t *testing.T, c *Client) {
-	if c.UserAgent != userAgent {
-		t.Errorf("NewClient UserAgent = %v, expected %v", c.UserAgent, userAgent)
+	if c.UserAgent != DefaultUserAgent() {
```
MihaiBojin, line 96: updated the test suite to expect the new UA

mongodbatlas/version.go
```diff
@@ -0,0 +1,15 @@
...
+
+const (
+	// UserAgent the default user agent string reported by this client
+	UserAgent = "go-mongodbatlas"
+	// ClientVersion the version of the current API client
+	ClientVersion = "0.4.0" // Should be set to the next version planned to be released
+)
+
+// DefaultUserAgent constructs the default user agent
+func DefaultUserAgent() string {
```
MihaiBojin, line 13: helper function to construct the default UA string

mongodbatlas/version.go
```diff
@@ -0,0 +1,15 @@
+package mongodbatlas
```
gssbzn, line 1: [non-blocking] I don't feel moving this to a version file adds any value vs the mongodbatlas. If we were to have a package `version` then moving _only_ the `ClientVersion` there would make more sense

mongodbatlas/version.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "fmt"
+
+const (
+	// UserAgent the default user agent string reported by this client
+	UserAgent = "go-mongodbatlas"
```
gssbzn, line 7: I think we should not export this
```suggestion
	userAgent = "go-mongodbatlas"
```

mongodbatlas/version.go
```diff
@@ -0,0 +1,15 @@
...
+	// ClientVersion the version of the current API client
+	ClientVersion = "0.4.0" // Should be set to the next version planned to be released
```
gssbzn, line 9: it's more common to just say `Version`
```suggestion
	// Version the version of the current API client
	Version = "0.4.0" // Should be set to the next version planned to be released
```

mongodbatlas/version.go
gssbzn, line 13: [nit] similar to the const I don't see a value on exporting this or even why we have 
`UserAgent` as a const and `DefaultUserAgent` as a func
we should just offer one, maybe 

```go
const libraryName = "go-mongodbatlas"
const Version = "0.4"
var userAgent = fmt.Sprintf("%s/%s (%s;%s)", libraryName, Version, runtime.GOOS, runtime.GOARCH)
```

mongodbatlas/version.go
MihaiBojin: version is very generic though; hard to search for

mongodbatlas/version.go
MihaiBojin: yeah fair enough, updated

mongodbatlas/version.go
MihaiBojin: still, updated since it's more go idiomatic

mongodbatlas/version.go
MihaiBojin: nice!

gssbzn:
in mcli we have a [RELEASING.md](https://github.com/mongodb/mongocli/blob/master/RELEASING.md) we also have a little bash script to automate some of the process, we could look if making a bash script to update the version and tag the release would make sense here

themantissa:
@gssbzn this is an excellent example, thank you!  Sounds like we'll want similar for this but perhaps figure out communication channels between teams for timing, etc.

.github/workflows/release-check.yml
```diff
@@ -0,0 +1,18 @@
...
+    name: Release sanity tests
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Checking code health
+        run: make tools lint test
+
+      - name: Checking the version
+        run: make check-version
```
MihaiBojin, line 18: This workflow will ensure that we get notified when a tag was pushed and the version was not updated.

GNUmakefile
```diff
@@ -13,26 +13,36 @@ link-git-hooks:
 	find .git/hooks -type l -exec rm {} \;
 	find githooks -type f -exec ln -sf ../../{} .git/hooks/ \;
 
+.PHONY: build
```
MihaiBojin, line 16: updated for consistency

README.md
```diff
@@ -67,6 +67,9 @@ Each version of the client is tagged and the version is updated accordingly.
 
 To see the list of past versions, run `git tag`.
 
+To release a new version, first ensure that `[Version](./mongodbatlas/mongodbatlas.go)' is updated
```
gssbzn, line 70: should 
>`[Version](./mongodbatlas/mongodbatlas.go)'
be
`[Version](./mongodbatlas/mongodbatlas.go)` 
also does the linking works like this? sometimes for file you need to add blob in the path

scripts/check-version.sh
```diff
@@ -0,0 +1,31 @@
...
+    echo "ERROR: Mismatch between the release version ('$tag') and '$VERSION_FILE'"
+    echo "$VERSION_FILE:"
+    echo "---------"
+    grep "Version = " "$VERSION_FILE"
+    echo "---------"
+    echo
+    exit 1
+fi
+echo "OK."
+echo
```
gssbzn, line 31: [super-nit] missing new line

README.md
MihaiBojin: nice catch on the backticks!
Yup, linking like that (against a file or dir, relative to the current file) works due to GitHub Markdown syntax; e.g.: https://github.com/mongodb/go-client-mongodb-atlas/blob/ed14165966a89e84a67e79ad7aeabedbab2734b9/README.md

PR #122, opened 2020-08-11
chore: add dependabot

gssbzn:
## Description

This adds github dependabot, we can use this to keep dependencies in check

## Type of change:

## Required Checklist:

## Further comments

Example PR https://github.com/mongodb/mongocli/pull/345

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: "/"
+    schedule:
+      interval: weekly
+      day: monday
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,5 @@ updates:
     schedule:
       interval: weekly
       day: monday
+    reviewers:
+      - "mongodb/dependabot-team"


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: "/"
+    schedule:
+      interval: weekly
+      day: monday
```
gssbzn, line 7: we can also auto assign people to the PRs but I think we should probably handle this with a group, @themantissa what do you think here?

themantissa: A group makes it much less likely to get missed if someone is OOO - perhaps just internal MDBers on your team? I can be in the group as well if you'd like to ensure someone is keeping an eye out?

gssbzn: yep, we can make this an internal devs group, I'm no admin here so just let me know the name of the group and I'll add it here

themantissa:
@gssbzn created @mongodb/dependabot-team and added you.  I'm assuming this is what you mean by group?  If so I can add others as needed.

gssbzn:
@themantissa FYI I did a run on this on mcli and you may need to invite the team to the repo

PR #125, opened 2020-08-17
CLOUDP-69426: Atlas go client - Update PeersService.List to use providerName

andreaangiolillo:
## Description

[Get All Network Peering Connections in A Project](https://docs.atlas.mongodb.com/reference/api/vpc-get-connections-list/#request-query-parameters): `providerName` is not set for the endpoint`GET /groups/{GROUP-ID}/peers`.

Link to any related issue(s): 

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/containers.go
+++ b/mongodbatlas/containers.go
@@ -13,7 +13,7 @@ const containersPath = "groups/%s/containers"
 //
 // See more: https://docs.atlas.mongodb.com/reference/api/vpc/
 type ContainersService interface {
-	List(context.Context, string, *NetworkPeeringListOptions) ([]Container, *Response, error)
+	List(context.Context, string, *ContainersListOptions) ([]Container, *Response, error)
 	ListAll(context.Context, string, *ListOptions) ([]Container, *Response, error)
 	Get(context.Context, string, string) (*Container, *Response, error)
 	Create(context.Context, string, *Container) (*Container, *Response, error)
@@ -27,7 +27,7 @@ type ContainersServiceOp service
 
 var _ ContainersService = &ContainersServiceOp{}
 
-type NetworkPeeringListOptions struct {
+type ContainersListOptions struct {
 	ProviderName string `url:"providerName,omitempty"`
 	ListOptions
 }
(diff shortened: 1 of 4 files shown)


mongodbatlas/containers.go
```diff
@@ -27,7 +27,7 @@ type ContainersServiceOp service
 
 var _ ContainersService = &ContainersServiceOp{}
 
-type ContainersListOptions struct {
```
gssbzn, line 30: Let's try to minimize the impact of this change by not renaming the struct, if a client is using this then we're are going to break their usage

gssbzn:
@themantissa this is a bit urgent as I'm realising that if `providerName` is not given it defaults to AWS so right now there's no way to list GCP or Azure peers

mongodbatlas/peers.go
```diff
@@ -13,7 +13,7 @@ const peersPath = "groups/%s/peers"
 // endpoints of the MongoDB Atlas API.
 // See more: https://docs.atlas.mongodb.com/reference/api/peers/
```
gssbzn, line 14: I have a small drive by fix here if you can address this here as well please
```suggestion
// See more: https://docs.atlas.mongodb.com/reference/api/vpc/
```

themantissa:
@gssbzn we ran into the same issue - which is why there is now a resource to get all (we couldn't change the original without breaking people's existing scripts) - did you all see https://docs.atlas.mongodb.com/reference/api/vpc-get-containers-list-all/?  I'll approve this but we'll need to make sure the way we handle it in terraform is not impacted here.  I'll add the DoU devs and ask them to review today.

gssbzn:
@themantissa the one you link is for Containers this is for `Peers`( Network peering) which has similar behaviour, as I can see and I could not find an equivalent like we have in containers to get all

gssbzn:
This won't break existing users in the sense that the result is going to be different, current behaviour remains, where calling this method with no params default to listing AWS only, this is whay it's a big bug for us where I wanted to test creating an Azure peering and I could not see it as there's no way for me to list Azure since I need the query param

TL;DR
`/peers` return AWS by default
if I want Azure (which I just tried as I'm creating an Azure peering) I need `/peers?providerName=Azure` which I can't do with the current client

themantissa:
Ah, that makes sense (I blame lack of coffee).  Being able to list containers was much more of a need for us since they can be created by more than just the actual container actions (first cluster creation, privatelink for example both create a container if one suiting it does not exist).  Not as much for peers but that makes sense for the CLI!

PR #137, opened 2020-09-14
CLOUDP-72746: Add Performance Advisor API call to atlas go client

andreaangiolillo:
## Description

Implemented Performance Advisor API call https://docs.atlas.mongodb.com/reference/api/performance-advisor/

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/performance_advisor.go
+++ b/mongodbatlas/performance_advisor.go
@@ -13,20 +13,20 @@ const (
 	performanceAdvisorSuggestedIndexesLogsPath = performanceAdvisorPath + "/suggestedIndexes"
 )
 
-// ProcessAdvisorService is an interface of the Process Advisor
+// PerformanceAdvisorService is an interface of the Performance Advisor
 // endpoints of the MongoDB Atlas API.
 //
 // See more: https://docs.atlas.mongodb.com/reference/api/performance-advisor/
-type ProcessAdvisorService interface {
+type PerformanceAdvisorService interface {
 	GetNamespaces(context.Context, string, string, *PerformanceAdvisorRequest) (*Namespaces, *Response, error)
 	GetSlowQueries(context.Context, string, string, *PerformanceAdvisorRequest) (*SlowQueries, *Response, error)
 	GetSuggestedIndexes(context.Context, string, string, *PerformanceAdvisorRequest) (*SuggestedIndexes, *Response, error)
 }
 
-// ProcessAdvisorService handles communication with the Process Advisor related methods of the MongoDB Atlas API
-type ProcessAdvisorServiceOp service
+// PerformanceAdvisorServiceOp handles communication with the Performance Advisor related methods of the MongoDB Atlas API
+type PerformanceAdvisorServiceOp service
 
-var _ ProcessAdvisorService = &ProcessAdvisorServiceOp{}
+var _ PerformanceAdvisorService = &PerformanceAdvisorServiceOp{}
 
 // Namespace represents a Namespace.
 type Namespace struct {
@@ -106,7 +106,7 @@ type PerformanceAdvisorRequest struct {
 // GetNamespaces retrieves the namespaces for collections experiencing slow queries for a specified host.
 //
 // See more: https://docs.atlas.mongodb.com/reference/api/pa-namespaces-get-all/
-func (s *ProcessAdvisorServiceOp) GetNamespaces(ctx context.Context, groupID, processName string, body *PerformanceAdvisorRequest) (*Namespaces, *Response, error) {
+func (s *PerformanceAdvisorServiceOp) GetNamespaces(ctx context.Context, groupID, processName string, body *PerformanceAdvisorRequest) (*Namespaces, *Response, error) {
 	if groupID == "" {
 		return nil, nil, NewArgError("groupID", "must be set")
 	}
@@ -138,7 +138,7 @@ func (s *ProcessAdvisorServiceOp) GetNamespaces(ctx context.Context, groupID, pr
 // GetSlowQueries gets log lines for slow queries as determined by the Performance Advisor.
 //
 // See more: https://docs.atlas.mongodb.com/reference/api/pa-get-slow-query-logs/
-func (s *ProcessAdvisorServiceOp) GetSlowQueries(ctx context.Context, groupID, processName string, body *PerformanceAdvisorRequest) (*SlowQueries, *Response, error) {
+func (s *PerformanceAdvisorServiceOp) GetSlowQueries(ctx context.Context, groupID, processName string, body *PerformanceAdvisorRequest) (*SlowQueries, *Response, error) {
 	if groupID == "" {
 		return nil, nil, NewArgError("groupID", "must be set")
 	}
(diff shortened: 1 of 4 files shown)


mongodbatlas/performance_advisor.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, nil, NewArgError("processName", "must be set")
+	}
+
+	if body == nil {
+		return nil, nil, NewArgError("body", "must be set")
+	}
+
+	path := fmt.Sprintf(performanceAdvisorNamespacesPath, groupID, processName)
+
+	req, err := s.Client.NewRequest(ctx, http.MethodGet, path, body)
```
gssbzn, line 124: [nit] couple of things `PerformanceAdvisorRequest` is not actually a body request, they are query params, as such I'd recommend you to rename this to `PerformanceAdvisorQuery`, also all query params by definition are optional and finally for future reference by definition GET methods can never have a body on their request and if by chance you manage to set one (some clients won't let you) servers usually ignore them

mongodbatlas/performance_advisor.go
```diff
@@ -0,0 +1,200 @@
...
+	Since      int64  `json:"since,omitempty"`      // Point in time, specified as milliseconds since the Unix Epoch, from which you want to receive results.
+	Duration   int64  `json:"duration,omitempty"`   // 	Length of time from the since parameter, in milliseconds, for which you want to receive results.
+	Envelope   bool   `json:"envelope,omitempty"`   // Boolean that specifies whether or not to wrap the response in an envelope. Default is false.
+	Namespaces string `json:"namespaces,omitempty"` // Namespaces from which to retrieve slow query logs. A namespace consists of the database and collection resource separated by a ., such as <database>.<collection>.
+	NLogs      int64  `json:"nLogs,omitempty"`      // Maximum number of log lines to return. Defaults to 20000.
+	Pretty     bool   `json:"pretty,omitempty"`     // Display response in a prettyprint format. Defaults to false.
+	NIndexes   int64  `json:"nIndexes,omitempty"`   // Maximum number of indexes to suggest. Defaults to unlimited.
+	NExamples  int64  `json:"NExamples,omitempty"`  // Maximum number of examples queries to provide that will be improved by a suggested index. Defaults to 5.
```
gssbzn, line 103: I already suggested renaming this to query or maybe options but I'd also split this into 3 structures given that not all fields are available to all three methods `NIndexes` and `NExamples` are Indexes only `nLogs` is slow query only, this may not be a problem as omit empty will solve not setting it but what if I set it? or if me as a user think I can set it just to be ignore? so my comment is more about usability than anything else, we should avoid users being confused by the usage of this struct and as such we should offer just what they can set

Finally given that this is for query params don't use the `json` tag, it should be `url`

gssbzn (changes requested):
Good start but one big thing here about passing the query params as part of the body

mongodbatlas/performance_advisor.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+// PerformanceAdvisorService is an interface of the Performance Advisor
+// endpoints of the MongoDB Atlas API.
+//
+// See more: https://docs.atlas.mongodb.com/reference/api/performance-advisor/
+type PerformanceAdvisorService interface {
+	GetNamespaces(context.Context, string, string, *NamespacesOptions) (*Namespaces, *Response, error)
+	GetSlowQueries(context.Context, string, string, *SlowQueriesOptions) (*SlowQueries, *Response, error)
+	GetSuggestedIndexes(context.Context, string, string, *SuggestedIndexesOptions) (*SuggestedIndexes, *Response, error)
```
gssbzn, line 23: [grammar] only the last word as plural
```suggestion
	GetNamespaces(context.Context, string, string, *NamespaceOptions) (*Namespaces, *Response, error)
	GetSlowQueries(context.Context, string, string, *SlowQueryOptions) (*SlowQueries, *Response, error)
	GetSuggestedIndexes(context.Context, string, string, *SuggestedIndexOptions) (*SuggestedIndexes, *Response, error)
```

mongodbatlas/performance_advisor.go
```diff
@@ -0,0 +1,230 @@
...
+	SuggestedIndexes []*SuggestedIndex `json:"suggestedIndexes,omitempty"` // Documents with information about the indexes suggested by the Performance Advisor.
+	Shapes           []*Shape          `json:"shapes,omitempty"`           // Documents with information about the query shapes that are served by the suggested indexes.
+
+}
+
+// SlowQueriesOptions contains the request query parameters for the API request.
+type SlowQueriesOptions struct {
+	Since      int64  `url:"since,omitempty"`      // Point in time, specified as milliseconds since the Unix Epoch, from which you want to receive results.
+	Duration   int64  `url:"duration,omitempty"`   // 	Length of time from the since parameter, in milliseconds, for which you want to receive results.
+	Envelope   bool   `url:"envelope,omitempty"`   // Boolean that specifies whether or not to wrap the response in an envelope. Default is false.
```
gssbzn, line 98: we don't support envelope in the client, this would change the response and we wouldn't be able de deserialized

mongodbatlas/performance_advisor.go
```diff
@@ -0,0 +1,230 @@
...
+	Namespaces string `url:"namespaces,omitempty"` // Namespaces from which to retrieve slow query logs. A namespace consists of the database and collection resource separated by a ., such as <database>.<collection>.
+	NLogs      int64  `url:"nLogs,omitempty"`      // Maximum number of log lines to return. Defaults to 20000.
+	Pretty     bool   `url:"pretty,omitempty"`     // Display response in a prettyprint format. Defaults to false.
```
gssbzn, line 101: similar to envelope this is not supported, mainly as it has no meaning in for the client, nothing would change

mongodbatlas/performance_advisor.go
```diff
@@ -0,0 +1,230 @@
...
+	Since    int64 `url:"since,omitempty"`    // Point in time, specified as milliseconds since the Unix Epoch, from which you want to receive results.
+	Duration int64 `url:"duration,omitempty"` // 	Length of time from the since parameter, in milliseconds, for which you want to receive results.
+	Envelope bool  `url:"envelope,omitempty"` // Boolean that specifies whether or not to wrap the response in an envelope. Default is false.
+	Pretty   bool  `url:"pretty,omitempty"`   // Display response in a prettyprint format. Defaults to false.
+}
+
+// SuggestedIndexesOptions contains the request query parameters for the API request.
+type SuggestedIndexesOptions struct {
+	Since      int64  `url:"since,omitempty"`      // Point in time, specified as milliseconds since the Unix Epoch, from which you want to receive results.
+	Duration   int64  `url:"duration,omitempty"`   // 	Length of time from the since parameter, in milliseconds, for which you want to receive results.
```
gssbzn, line 115: [non-blocking] you could've embedded the  `NamespacesOptions` here and in `SlowQueriesOptions` but not needed

mongodbatlas/performance_advisor.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, nil, NewArgError("groupID", "must be set")
+	}
+
+	if processName == "" {
+		return nil, nil, NewArgError("processName", "must be set")
+	}
+
+	if opts == nil {
+		return nil, nil, NewArgError("opts", "must be set")
+	}
```
gssbzn, line 138: Options are optional so this is safe to be  nil

gssbzn (changes requested):
Almost there but couple of small things
1, we don't support envelope or pretty as query params
2, options are optional so it should be safe to pass nil so we should remove those validations

gssbzn (approved):
@themantissa ready for you

gssbzn:
@andreaangiolillo unrelated to your change cna you do `go mod tidy`? I think there are some differences now and I don't want to do a separate PR just to fix the checksums

mongodbatlas/performance_advisor_test.go
```diff
@@ -0,0 +1,237 @@
...
+					"namespace" : "data.zips",
+					"type" : "COLLECTION"
+				  }, {
+					"namespace" : "data.stocks",
+					"type" : "COLLECTION"
+				  } ]
+		}`)
+	})
+
+	body := &NamespaceOptions{
```
gssbzn, line 31: checking the Ops Manager PR I just saw this is still referenced as body in our tests, let's fix this

gssbzn (changes requested):

gssbzn (approved):
LGTM

themantissa (approved):
LGTM!   Don't see any near term downstream work on our side so feel free to proceed.  Ty!

Issue #143, opened 2020-10-01
Add "Name" ListOption for filtering list results


sunny-b:
The Mongo Atlas API supports filtering list results based on a `name` parameter.

Link: https://docs.atlas.mongodb.com/reference/api/organization-get-all/#resource

It would be really useful to have this supported in the Go Client as well. At the moment, it looks like the only ListOptions that are supported are `PageNum`, and `ItemsPerPage` https://github.com/mongodb/go-client-mongodb-atlas/blob/a663e096685c8ae03b9970767287550593f2e6ba/mongodbatlas/mongodbatlas.go#L130-L138

We use Ops Manager at work and we depend on this feature in the API. We want to use this Go client but we would need this feature to be supported.

It is a pretty small change and I could easily make a PR for it. I've already made the change in my own fork:
https://github.com/sunny-b/go-client-mongodb-atlas/commit/c581130f09a4a26c3a538c0777d4f768cce96c7b

sunny-b:
Ah I see it might be better to create an `OrganizationListOptions` struct similar to `ContainerListOptions` https://github.com/mongodb/go-client-mongodb-atlas/blob/d1f10b2b037bf8a783bada68f9c40f9d56a05a1f/mongodbatlas/containers.go#L30-L33

sunny-b:
I'm closing this as it was fixed by #144

PR #144, opened 2020-10-01
add OrganizationsListOptions for filtering organization list results

sunny-b:
## Description

Addresses #143 

The Ops Manager and Atlas API supports filtering Organization list results by the name, however this Go client does not support that at the moment. This PR would add that functionality and adds a test to help ensure it works.

I recognize this is a breaking change to the interface so I'm open to discussion on a better way to do it. Though I saw this pattern in `ContainersListOptions` which is where I adopted it from.

https://github.com/mongodb/go-client-mongodb-atlas/blob/d1f10b2b037bf8a783bada68f9c40f9d56a05a1f/mongodbatlas/containers.go#L30-L33

Link to any related issue(s): 

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/mongodbatlas_test.go
+++ b/mongodbatlas/mongodbatlas_test.go
@@ -494,3 +494,55 @@ func checkCurrentPage(t *testing.T, resp *Response, expectedPage int) { //nolint
 		t.Fatalf("expected current page to be '%d', was '%d'", expectedPage, p)
 	}
 }
+
+func TestSetListOptions_EmptyBoolPtr(t *testing.T) {
+	type testListOptions struct {
+		Exists *bool `url:"exists,omitempty"`
+	}
+
+	boolPtr := func(b bool) *bool {
+		return &b
+	}
+	testBaseUrl := "www.example.com"
+
+	tests := []struct {
+		testName string
+		arg      *testListOptions
+		expected string
+	}{
+		{
+			testName: "true value",
+			arg: &testListOptions{
+				Exists: boolPtr(true),
+			},
+			expected: testBaseUrl + "?exists=true",
+		},
+		{
+			testName: "false value",
+			arg: &testListOptions{
+				Exists: boolPtr(false),
+			},
+			expected: testBaseUrl + "?exists=false",
+		},
+		{
+			testName: "omitted null value",
+			arg:      &testListOptions{},
+			expected: testBaseUrl,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			t.Helper()
+			t.Log(tt.testName)
+
+			actual, err := setListOptions(testBaseUrl, tt.arg)
+			if err != nil {
+				t.Errorf("setListOptions() error = %v", err)
+			}
+			if actual != tt.expected {
+				t.Errorf("setListOptions() actual = %s, expected = %s", actual, tt.expected)
+			}
+		})
+	}
+}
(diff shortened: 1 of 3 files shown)


mongodbatlas/organizations.go
```diff
@@ -40,6 +40,12 @@ type OrganizationsServiceOp service
 
 var _ OrganizationsService = &OrganizationsServiceOp{}
 
+// OrganizationsListOptions filtering options for organizations
+type OrganizationsListOptions struct {
+	Name string `url:"name,omitempty"`
```
gssbzn, line 45: Since you mention Ops Manager, Ops manager also has a boolean, `includeDeletedOrgs`, I've been wanting to fix this Listing option to add that one in particular but since we are breaking this I wouldn't mind taking the extra step here.
If you decide to add it be aware that it defaults to true so `omitempty` may get tricky

gssbzn (commented):
Hi @sunny-b, thanks for your contribution, I'm happy to break the client to fix this, just one small comment in case you want to improve OM support as well.
I'll let @themantissa have final say on this breaking change

mongodbatlas/organizations.go
sunny-b: I think using a bool pointer can help get around the `omitempty` business. According to the docs, if the `includeDeletedOrgs` param is missing in the url, then it'll still default to `true`.

image


I added in that change and made a test to show that `setListOptions` will properly omit an empty bool pointer. Please let me know if you need anything else for this PR.

gssbzn: thanks for adding this, yes, making it a pointer should work

mongodbatlas/mongodbatlas_test.go
```diff
@@ -494,3 +494,55 @@ func checkCurrentPage(t *testing.T, resp *Response, expectedPage int) { //nolint
...
+			expected: testBaseURL,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			t.Helper()
+			t.Log(tt.testName)
+
+			actual, err := setListOptions(testBaseURL, tt.arg)
```
gssbzn, line 539: there's another linting error on this line, go thing with pointers and functions inside loops

themantissa (dismissed):
LGTM as I don't think this impacts us particularly.  However added the DoU devs for a quick review to check just in case. FYI @coderGo93 and @leofigy Thanks all!

gssbzn (approved):

sunny-b:
Anyway of getting the appropriate reviews for this? I would love for this feature to get merged in. We're planning on using this Go client at work and this is a much needed feature for us.

leofigy (approved):
LGTM 👍🏽

gssbzn:
@sunny-b feel free to add update https://github.com/mongodb/go-client-mongodb-ops-manager with this change as well, we may be doing this anyway in the coming days

sunny-b:
@gssbzn Yeah that's a good idea. Though you'd need to put out a new release of this package in order to update Ops Manager as it uses the various `ListOptions` from this package.

gssbzn:
@sunny-b on the ops-manager-client we usually update to master (`go get -u go.mongodb.org/atlas@master`), not ideal but we are ok with it since we manage both packages. releasing a new version for the atlas client takes a bit more time as we coordinate this with the team in charge of the terraform provider. Don't feel any pressure on doing this, we'll address this on our next update of the ops manager client dependency for atlas but I can't promise any dates right

sunny-b:
@gssbzn https://github.com/mongodb/go-client-mongodb-ops-manager/pull/59

PR #165, opened 2021-01-08
INTMDB-157: Add deprecation notice

themantissa:
## Description

Add deprecation notice to both programmatic key whitelist and project ip whitelist, replaced with access list.

Link to any related issue(s): 

## Type of change:

Adds simple deprecation notice in code

## Required Checklist:

### DIFF
--- a/mongodbatlas/project_ip_whitelist.go
+++ b/mongodbatlas/project_ip_whitelist.go
@@ -7,6 +7,13 @@ import (
 	"net/url"
 )
 
+/*  IMPORTANT: DEPRECATION NOTICE
+Access List Replaces Whitelist
+Atlas now refers to its cluster firewall management as IP Access Lists. 
+Atlas has deprecated the whitelist resource and will disable it in June 2021. 
+Please update any dependent work to use project_ip_access_list.go 
+*/
+
 const projectIPWhitelistPath = "groups/%s/whitelist"
 
 // ProjectIPWhitelistService is an interface for interfacing with the Project IP Whitelist
--- a/mongodbatlas/project_ip_whitelist.go
+++ b/mongodbatlas/project_ip_whitelist.go
@@ -8,7 +8,7 @@ import (
 )
 
 /* IMPORTANT: DEPRECATION NOTICE
-Access List Replaces Whitelist
+Deprecated: Access List Replaces Whitelist
 Atlas now refers to its cluster firewall management as IP Access Lists.
 Atlas has deprecated the whitelist resource and will disable it in June 2021.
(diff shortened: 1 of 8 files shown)


mongodbatlas/whitelist_api_keys.go
```diff
@@ -6,6 +6,13 @@ import (
 	"net/http"
 )
 
+/* IMPORTANT: DEPRECATION NOTICE
+Access List Replaces Whitelist
+Atlas now refers to programmatic API key whitelists as access lists.
+Atlas has deprecated the whitelist method and will disable it in June 2021.
+Please update any dependent work to use accesslist_api_keys.go
+*/
```
jamesbroadhead, line 14: I think if we follow this format https://github.com/golang/go/wiki/Deprecated then tools will pick up the fact that this is a deprecation announcement automatically

jamesbroadhead (dismissed):

mongodbatlas/whitelist_api_keys.go
themantissa: Thank you so much @jamesbroadhead, did not know that!   Added commit to fit the spec.

mongodbatlas/project_ip_whitelist.go
```diff
@@ -7,6 +7,13 @@ import (
 	"net/url"
 )
 
+/* IMPORTANT: DEPRECATION NOTICE
+Deprecated: Access List Replaces Whitelist
```
gssbzn, line 11: [super-nit] can we move this below line 21? james cover how to do this in go but right now making this a global comment makes the whole package (`mongodbatlas`) deprecated, you want to flag that only the `ProjectIPWhitelistService` is the one being deprecated 

Same with `WhitelistAPIKeysService` and I think we can skip commenting on the test file

gssbzn (changes requested):

mongodbatlas/project_ip_whitelist.go
themantissa: Removed from tests and moved lower in latest commit.

gssbzn (approved):
Sorry, pointing to the line number is too complex so I pushed something here to avoid going in circles, also this kind of works
Screenshot 2021-01-11 at 18 45 17

LGTM

themantissa:
@gssbzn sometimes it's easier to do than explain - thank you :)   @jamesbroadhead  if you are good I'll merge this tomorrow so we can release (or you can, either works)

jamesbroadhead:
oh yep - thanks for testing it out @gssbzn !

jamesbroadhead (approved):

PR #188, opened 2021-03-15
CLOUDP-83026: support for cluster status endpoint

antonlisovenko:
## Description

This PR adds support for https://docs.atlas.mongodb.com/reference/api/clusters-check-operation-status/ endpoint to clusters.

Link to any related issue(s): 

## Type of change:

- [x] This change requires a documentation update

## Required Checklist:

### DIFF
--- a/mongodbatlas/clusters.go
+++ b/mongodbatlas/clusters.go
@@ -21,19 +21,27 @@ import (
 	"net/url"
 )
 
+type ChangeStatus string
+
+const (
+	ChangeStatusApplied ChangeStatus = "APPLIED"
+	ChangeStatusPending ChangeStatus = "PENDING"
+)
+
 const clustersPath = "groups/%s/clusters"
 
 // ClustersService is an interface for interfacing with the Clusters
 // endpoints of the MongoDB Atlas API.
 // See more: https://docs.atlas.mongodb.com/reference/api/clusters/
 type ClustersService interface {
-	List(context.Context, string, *ListOptions) ([]Cluster, *Response, error)
-	Get(context.Context, string, string) (*Cluster, *Response, error)
-	Create(context.Context, string, *Cluster) (*Cluster, *Response, error)
-	Update(context.Context, string, string, *Cluster) (*Cluster, *Response, error)
-	Delete(context.Context, string, string) (*Response, error)
-	UpdateProcessArgs(context.Context, string, string, *ProcessArgs) (*ProcessArgs, *Response, error)
-	GetProcessArgs(context.Context, string, string) (*ProcessArgs, *Response, error)
+	List(ctx context.Context, groupID string, options *ListOptions) ([]Cluster, *Response, error)
+	Get(ctx context.Context, groupID, clusterName string) (*Cluster, *Response, error)
+	Create(ctx context.Context, groupID string, cluster *Cluster) (*Cluster, *Response, error)
+	Update(ctx context.Context, groupID, clusterName string, cluster *Cluster) (*Cluster, *Response, error)
+	Delete(ctx context.Context, groupID, clusterName string) (*Response, error)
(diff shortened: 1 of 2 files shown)


mongodbatlas/clusters.go
```diff
@@ -21,19 +21,27 @@ import (
...
-	UpdateProcessArgs(context.Context, string, string, *ProcessArgs) (*ProcessArgs, *Response, error)
-	GetProcessArgs(context.Context, string, string) (*ProcessArgs, *Response, error)
+	List(ctx context.Context, groupID string, options *ListOptions) ([]Cluster, *Response, error)
+	Get(ctx context.Context, groupID, clusterName string) (*Cluster, *Response, error)
+	Create(ctx context.Context, groupID string, cluster *Cluster) (*Cluster, *Response, error)
+	Update(ctx context.Context, groupID, clusterName string, cluster *Cluster) (*Cluster, *Response, error)
+	Delete(ctx context.Context, groupID, clusterName string) (*Response, error)
+	UpdateProcessArgs(ctx context.Context, groupID, clusterName string, args *ProcessArgs) (*ProcessArgs, *Response, error)
+	GetProcessArgs(ctx context.Context, groupID, clusterName string) (*ProcessArgs, *Response, error)
+	Status(ctx context.Context, groupID, clusterName string) (ClusterStatus, *Response, error)
```
antonlisovenko, line 44: changed the signatures to include the names of parameters - this allows to distinguish two `string` parameters easier while working with the library. Found myself going to implementations regularly to make sure what parameters mean.
This is not critical and I can revert this in case more concise signatures are preferable!

mongodbatlas/clusters.go
```diff
@@ -395,6 +408,31 @@ func (s *ClustersServiceOp) GetProcessArgs(ctx context.Context, groupID, cluster
 	return root, resp, err
 }
 
+// Status gets the status of the operation on the Cluster.
+// See more: https://docs.atlas.mongodb.com/reference/api/clusters-check-operation-status/
```
gssbzn, line 412: [nit] I have some pending work on fixing the other methods but this renders the wrong way on the go docs so let's not add to the noise
```suggestion
// Status gets the status of the operation on the Cluster.
//
// See more: https://docs.atlas.mongodb.com/reference/api/clusters-check-operation-status/
```

mongodbatlas/clusters.go
```diff
@@ -395,6 +408,31 @@ func (s *ClustersServiceOp) GetProcessArgs(ctx context.Context, groupID, cluster
...
+		return ClusterStatus{}, nil, err
+	}
+
+	root := new(ClusterStatus)
+	resp, err := s.Client.Do(ctx, req, root)
+	if err != nil {
+		return ClusterStatus{}, resp, err
+	}
+
+	return *root, resp, err
```
gssbzn, line 433: [nit] I guess you don't want to deal with pointers (not going to judge on that) but you could reuse the empty object a bit more like
```suggestion
	var root ClusterStatus
	req, err := s.Client.NewRequest(ctx, http.MethodGet, path, nil)
	if err != nil {
		return root, nil, err
	}

	resp, err := s.Client.Do(ctx, req, &root)
	return root, resp, err
```

would that work for your case (not blocking on this if it doesn't work for what you're going here)?

gssbzn (changes requested):
One small nit on the doc style (blocking) and one recommendation/question(not blocking) thanks for the addition here

mongodbatlas/clusters.go
antonlisovenko: yes, this looks better, thanks! 👍

gssbzn (approved):
LGTM, welcome to the go client code 🎉

gssbzn:
@themantissa FYI

PR #221, opened 2021-06-22
lint: enable godot

gssbzn:
## Description

Enable `godot` to improve our docs style


colm-quinn (approved):
That's going to be fun reminder everytime we write a doc string :) LGTM

gssbzn:
> That's going to be fun reminder everytime we write a doc string

run `make lint-fix` or `make check` (test and fix lint, also enabled on the pre commit), this lint has an auto fix, this PR was done with it

PR #222, opened 2021-06-22
lint: replace golint with revive

gssbzn:
## Description

`golint` has been archived and it's now deprecated on golangci-lint, they recommend to replace with revive

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,12 +10,39 @@ linters-settings:
       - octalLiteral
   gocyclo:
     min-complexity: 15
-  golint:
-    min-confidence: 0
   govet:
     check-shadowing: true
   maligned:
     suggest-new: true
+  revive:
+    # see https://github.com/mgechev/revive#available-rules for details.
+    ignore-generated-header: true
+    severity: warning
+    rules:
+      - name: blank-imports
+      - name: context-as-argument
+      - name: context-keys-type
+      - name: dot-imports
+      - name: error-return
+      - name: error-strings
+      - name: error-naming
+      - name: exported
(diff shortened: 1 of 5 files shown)


.golangci.yml
```diff
@@ -51,6 +77,7 @@ linters:
     - misspell
     - nakedret
     - nolintlint
+    - revive # Fast, configurable, extensible, flexible, and beautiful linter for Go. Drop-in replacement of golint. [fast: false, auto-fix: false]
```
threebee, line 80: thanks for adding this comment

mongodbatlas/mongodbatlas.go
```diff
@@ -343,21 +344,7 @@ func SetUserAgent(ua string) ClientOpt {
...
-		return nil, err
-	}
-	var buf io.Reader
-	if body != nil {
-		if buf, err = c.newEncodedBody(body); err != nil {
-			return nil, err
-		}
-	}
-
-	req, err := http.NewRequest(method, u.String(), buf)
```
threebee, line 360: [q] For my own understanding, why was this removed?

threebee (approved):
LGTM, I have added a question.

mongodbatlas/mongodbatlas.go
gssbzn: moved to line 408-414

gssbzn: also for context, the linter was complaining `ctx` was unused, this is an exported method so changing the signature is frowned upon so better to use it, same for `NewGZipRequest` and `NewPlainRequest`

PR #223, opened 2021-06-23
INTMDB-17: Update search

abner-dou:
## Description

Added properties to search index and search analyzer struct in order to complain with the current mongodb atlas API documentation

Link to any related issue(s): 

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/search.go
+++ b/mongodbatlas/search.go
@@ -242,13 +242,15 @@ func (s *SearchServiceOp) UpdateAllAnalyzers(ctx context.Context, groupID, clust
 
 // SearchIndex index definition.
 type SearchIndex struct {
-	Analyzer       string        `json:"analyzer,omitempty"`
-	CollectionName string        `json:"collectionName"`
-	Database       string        `json:"database"`
-	IndexID        string        `json:"indexID,omitempty"`
-	Mappings       *IndexMapping `json:"mappings,omitempty"`
-	Name           string        `json:"name"`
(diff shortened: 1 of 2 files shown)


threebee (approved):
LGTM!

mongodbatlas/search.go
```diff
@@ -281,7 +283,8 @@ type IndexField struct {
 // SearchAnalyzer custom analyzer definition.
 type SearchAnalyzer struct {
 	BaseAnalyzer     string   `json:"baseAnalyzer"`
-	MaxTokenLength   *float64 `json:"maxTokenLength,omitempty"`
+	MaxTokenLength   *int     `json:"maxTokenLength,omitempty"`
```
gssbzn, line 286: any reason for this change? it would be better if we don't but not a blocker

themantissa (approved):
LGTM but first needs Gustavo's question answered on the reason for the float to int change.

mongodbatlas/search.go
abner-dou: According to the API documentation for search kndexes and analyzers, this property is an int type.

And that makes more sense (how you can limit a token using a float value?)

What do you think?

gssbzn: not a blocker like I said but it's a breaking change so trying to avoid it if we can, since for json all number fields are technically float, the previous implementation was valid (even if wrong) but feel free to merge

abner-dou: ok, It's good to know, so this time I'll merge that change

Thanks for your comment Gustavo

PR #233, opened 2021-07-05
Fix project invitation update

andreaangiolillo:
## Description

Bug: The Request Body Parameters ID, OrgID, GroupID must be blank when using update and updateByID endpoints for invitation management.

Link to any related issue(s): 

## Type of change:

- [x] Bug fix (non-breaking change which fixes an issue)

## Required Checklist:

### DIFF
--- a/mongodbatlas/project_invitations.go
+++ b/mongodbatlas/project_invitations.go
@@ -104,19 +104,19 @@ func (s *ProjectsServiceOp) InviteUser(ctx context.Context, invitation *Invitati
 // UpdateInvitation updates one pending invitation to the Atlas project that you specify.
 //
 // See more: https://docs.atlas.mongodb.com/reference/api/project-update-one-invitation/
-func (s *ProjectsServiceOp) UpdateInvitation(ctx context.Context, invitation *Invitation) (*Invitation, *Response, error) {
-	if invitation.GroupID == "" {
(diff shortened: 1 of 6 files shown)


fmenezes (approved):

mongodbatlas/project_invitations.go
```diff
@@ -124,8 +124,6 @@ func (s *ProjectsServiceOp) UpdateInvitationByID(ctx context.Context, groupID, i
 		return nil, nil, NewArgError("invitationID", "must be set")
 	}
 
-	invitation.ID = invitationID
-
```
colm-quinn, line 128: [q] If the parameters must be blank, should we instead check if they're set and reject on the call?

andreaangiolillo: The parameter `ID` inside the struct `invitation` (request body) must be black not the parameter itself. Here I was erroneously setting this param inside the body instead of using it just as a path parameter. See [Update One Project Invitation by Invitation ID](https://docs.atlas.mongodb.com/reference/api/project-update-one-invitation-by-id/).
To summarise, we are not ignoring the `ID` but we are just removing it from the request body and using it only in the path ([here I am using invitationID](https://github.com/mongodb/go-client-mongodb-atlas/blob/7103b52018f4586ce9fd620d0f99e5cc720207da/mongodbatlas/project_invitations.go#L159))

PR #236, opened 2021-07-23
CLOUDP-96138: Add serverless to the atlas go client

andreaangiolillo:
## Description

Jira Ticket: [CLOUDP-96138](https://jira.mongodb.org/browse/CLOUDP-96138)

Link to any related issue(s): 

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/clusters.go
+++ b/mongodbatlas/clusters.go
@@ -153,6 +153,7 @@ type Cluster struct {
 	MongoURIUpdated          string                   `json:"mongoURIUpdated,omitempty"`
 	MongoURIWithOptions      string                   `json:"mongoURIWithOptions,omitempty"`
 	Name                     string                   `json:"name,omitempty"`
+	CreateDate               string                   `json:"createDate,omitempty"`
 	NumShards                *int64                   `json:"numShards,omitempty"`
 	Paused                   *bool                    `json:"paused,omitempty"`
 	PitEnabled               *bool                    `json:"pitEnabled,omitempty"`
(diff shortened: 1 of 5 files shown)


mongodbatlas/clusters.go
```diff
@@ -153,6 +153,7 @@ type Cluster struct {
 	MongoURIUpdated          string                   `json:"mongoURIUpdated,omitempty"`
 	MongoURIWithOptions      string                   `json:"mongoURIWithOptions,omitempty"`
 	Name                     string                   `json:"name,omitempty"`
+	CreateDate               string                   `json:"createDate,omitempty"`
```
gssbzn, line 156: [q] are these and `Links` serverless only? if so can we add a godoc saying so?

andreaangiolillo: yes, they are. sure

gssbzn: I would also be in favour of maybe duplicating the create params with what you only need for create, it could be confusing for users to understand what params are dedicated clusters only and which ones are not

andreaangiolillo: no, sorry. They are also used with normal clusters

andreaangiolillo: `Links` is in the example of response for GET CLUSTER https://docs.atlas.mongodb.com/reference/api/clusters-get-one/

Screenshot 2021-07-23 at 14 48 39

gssbzn (commented):
Since I commented in the same thread probably got lost but...

I would also be in favour of maybe duplicating the create params with what you only need for create, it could be confusing for users to understand what params are dedicated clusters only and which ones are not, also it may be hard to maintain

gssbzn (approved):
LGTM thanks for the changes

PR #244, opened 2021-08-25
CLOUDP-98424: Allow upload of json in atlas search

fmenezes:
## Description

The team that owns atlas search suggested uploading JSON data instead of passing flags in our CLI experience to be inline with their efforts in the UI.

Link to any related issue(s): [CLOUDP-98424](https://jira.mongodb.org/browse/CLOUDP-98424)

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/search.go
+++ b/mongodbatlas/search.go
@@ -31,6 +31,7 @@ type SearchService interface {
 	ListIndexes(ctx context.Context, groupID string, clusterName string, databaseName string, collectionName string, opts *ListOptions) ([]*SearchIndex, *Response, error)
 	GetIndex(ctx context.Context, groupID, clusterName, indexID string) (*SearchIndex, *Response, error)
 	CreateIndex(ctx context.Context, projectID, clusterName string, r *SearchIndex) (*SearchIndex, *Response, error)
+	CreateIndexWithJson(ctx context.Context, projectID, clusterName string, body interface{}) (*SearchIndex, *Response, error)
 	UpdateIndex(ctx context.Context, projectID, clusterName, indexID string, r *SearchIndex) (*SearchIndex, *Response, error)
 	DeleteIndex(ctx context.Context, projectID, clusterName, indexID string) (*Response, error)
 	ListAnalyzers(ctx context.Context, groupID, clusterName string, listOptions *ListOptions) ([]*SearchAnalyzer, *Response, error)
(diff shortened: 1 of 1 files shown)


mongodbatlas/search.go
```diff
@@ -31,6 +31,7 @@ type SearchService interface {
 	ListIndexes(ctx context.Context, groupID string, clusterName string, databaseName string, collectionName string, opts *ListOptions) ([]*SearchIndex, *Response, error)
 	GetIndex(ctx context.Context, groupID, clusterName, indexID string) (*SearchIndex, *Response, error)
 	CreateIndex(ctx context.Context, projectID, clusterName string, r *SearchIndex) (*SearchIndex, *Response, error)
+	CreateIndexWithJSON(ctx context.Context, projectID, clusterName string, body interface{}) (*SearchIndex, *Response, error)
```
fmenezes, line 34: naming is hard 🤦

fmenezes: For context. the variable body is anything that can be converted to JSON, for instance `map[string]interface{}` or even `*SearchIndex`

gssbzn (commented):
do we really need this PR? I'm guessing is to allow search indexes with json files but clusters create already does that without such a method, see https://github.com/mongodb/mongocli/blob/ae1e135fe3b5f9f8500a5143169de2d0d7887ddb/internal/cli/atlas/clusters/create.go#L85-L89

mongodbatlas/search.go
colm-quinn: [q] Is it dangerous to just take raw json in here, rather than flags? I'm thinking of the potential to inject or send in other undesired data.

Also doesn't this make validation of the request harder? JSON comes in and all we end up saying is "it's failed, sorry"

fmenezes:
Thanks for your comments @colm-quinn @gssbzn, for context search team wants us to not parse the json and send raw data, much like their UI. Following @gssbzn example we could simple read data and convert to our struct but that would go not 100% inline with their request. Any thoughts?

fmenezes:
After some discussion internally we've decided to stop parsing part of the request as opposed to all of it.

@gssbzn this should be ready for another look.

gssbzn (approved):
LGTM on changes, I don't see support for synonyms but feel free to open a separate PR for those (we are suppose to add it soon)

cc @themantissa this is a breaking change for terraform I believe

PR #250, opened 2021-09-04
INTMDB-252: Added two parameters for Process args

coderGo93:
## Description

- Added two parameters `defaultReadConcern` and `defaultWriteConcern` for process args

Link to any related issue(s): 

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/clusters.go
+++ b/mongodbatlas/clusters.go
@@ -170,6 +170,8 @@ type Cluster struct {
 
 // ProcessArgs represents the advanced configuration options for the cluster.
 type ProcessArgs struct {
+	DefaultReadConcern               string `json:"defaultReadConcern,omitempty"`
+	DefaultWriteConcern              string `json:"defaultWriteConcern,omitempty"`
 	FailIndexKeyTooLong              *bool  `json:"failIndexKeyTooLong,omitempty"`
 	JavascriptEnabled                *bool  `json:"javascriptEnabled,omitempty"`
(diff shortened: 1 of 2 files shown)


mongodbatlas/clusters.go
```diff
@@ -170,6 +170,8 @@ type Cluster struct {
 
 // ProcessArgs represents the advanced configuration options for the cluster.
 type ProcessArgs struct {
+	DefaultReadConcern               string `json:"defaultReadConcern,omitempty"`
+	DefaultWriteConcern              string `json:"defaultWriteConcern,omitempty"`
```
gssbzn, line 174: [q] have you tested with a number as string (`"1"`)? reasons I ask is that technically this is "majority" (string) or `1`...`n` as number, API seems to accept string but want to make sure it's handled correctly, I'll merge after confirming this

gssbzn (approved):
LGTM with a question, will merge after confirming

andreaangiolillo (approved):
LGTM

mongodbatlas/clusters.go
coderGo93: I have not tested `defaultWriteConcern` as number, since the docs says is a string, will test `defaultWriteConcern` as number to see what happens

coderGo93: @gssbzn , I changed to int64 and it seems you can pass as integer/number in request, but the issue is in the response, it will panic like this  
`
 json: cannot unmarshal string into Go struct field ProcessArgs.defaultWriteConcern of type int64
`

 so in response it seems the `defaultWriteConcern` will always return as string

PR #274, opened 2022-01-24
CLOUDP-110641: Add device grant to atlas client

gssbzn:
## Description

Please include a summary of the fix/feature/change, including any relevant motivation and context.

Link to any related issue(s): 

## Type of change:

## Required Checklist:

### DIFF
--- a/auth/device_flow.go
+++ b/auth/device_flow.go
@@ -1,0 +1,99 @@
+// Copyright 2022 MongoDB Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	atlas "go.mongodb.org/atlas/mongodbatlas"
+)
+
+// DeviceCode holds information about the authorization-in-progress.
+type DeviceCode struct {
+	UserCode        string `json:"user_code"`        // UserCode is the code presented to users
+	VerificationURI string `json:"verification_uri"` // VerificationURI is the URI where users will need to confirm the code
+	DeviceCode      string `json:"device_code"`      // DeviceCode is the internal code to confirm the status of the flow
+	ExpiresIn       int    `json:"expires_in"`       // ExpiresIn when the code will expire
+	Interval        int    `json:"interval"`         // Interval who often to verify the status of the code
+
+	timeNow   func() time.Time
+	timeSleep func(time.Duration)
+}
+
+// RequestCode initiates the authorization flow by requesting a code.
+func (c Config) RequestCode(ctx context.Context) (*DeviceCode, *atlas.Response, error) {
+	req, err := c.NewRequest(ctx, http.MethodPost, "api/private/unauth/account/device/authorize",
+		url.Values{
+			"client_id": {c.ClientID},
+			"scope":     {strings.Join(c.Scopes, " ")},
+		},
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+	var r *DeviceCode
+	resp, err2 := c.Do(ctx, req, &r)
+	return r, resp, err2
+}
+
+// ErrTimeout is thrown when polling the server for the granted token has timed out.
+var ErrTimeout = errors.New("authentication timed out")
+
+// PollToken polls the server until an access token is granted or denied.
+func (c Config) PollToken(ctx context.Context, code *DeviceCode) (*Token, *atlas.Response, error) {
+	timeNow := code.timeNow
+	if timeNow == nil {
+		timeNow = time.Now
+	}
+	timeSleep := code.timeSleep
+	if timeSleep == nil {
+		timeSleep = time.Sleep
+	}
+
+	checkInterval := time.Duration(code.Interval) * time.Second
+	expiresAt := timeNow().Add(time.Duration(code.ExpiresIn) * time.Second)
+	req, err := c.NewRequest(ctx, http.MethodPost, "api/private/unauth/account/device/token",
+		url.Values{
+			"client_id":   {c.ClientID},
+			"device_code": {code.DeviceCode},
+			"grant_type":  {"urn:ietf:params:oauth:grant-type:device_code"},
+		},
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+	for {
+		timeSleep(checkInterval)
+		var r *Token
+		resp, err2 := c.Do(ctx, req, &r)
+		var target *atlas.ErrorResponse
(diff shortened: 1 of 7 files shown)


auth/device_flow.go
```diff
@@ -0,0 +1,99 @@
...
+	"context"
+	"errors"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	atlas "go.mongodb.org/atlas/mongodbatlas"
+)
+
```
andreaangiolillo, line 27: nit  `const deviceFlowBasePath = "api/private/unauth/account/device/"`

auth/device_flow.go
```diff
@@ -0,0 +1,99 @@
...
+		if err2 != nil {
+			return nil, resp, err2
+		}
+
+		if timeNow().After(expiresAt) {
+			return nil, nil, ErrTimeout
+		}
+		return r, resp, nil
+	}
+}
```
andreaangiolillo, line 99: [not blocking] I'd refactor this function as it has more than one responsibility (setting up the request + loop logic)

auth/oauth.go
```diff
@@ -0,0 +1,197 @@
...
+	// Token must be safe for concurrent use by multiple goroutines.
+	// The returned Token must not be modified.
+	Token() (*Token, error)
+}
+
+func (c *Config) Do(ctx context.Context, req *http.Request, v interface{}) (*atlas.Response, error) {
+	resp, err := atlas.DoRequestWithClient(ctx, c.client, req)
+	if err != nil {
+		// If we got an error, and the context has been canceled,
+		// the context's error is probably more useful.
```
andreaangiolillo, line 120: [q] When this scenario could happen?

auth/oauth.go
```diff
@@ -0,0 +1,197 @@
...
+		return nil, err
+	}
+
+	defer func() {
+		// Ensure the response body is fully read and closed
+		// before we reconnect, so that we reuse the same TCP connection.
+		// Close the previous response's body. But read at least some of
+		// the body so if it's small the underlying TCP connection will be
+		// re-used. No need to check for errors: if it fails, the Transport
+		// won't reuse it anyway.
```
andreaangiolillo, line 136: [q] Can we use [Transport](https://pkg.go.dev/net/http#Transport) for this client instead? The doc says `By default, Transport caches connections for future re-use.` Here is an example of how to use it https://go.dev/play/p/9Ah_lyfYxgV

auth/token.go
```diff
@@ -0,0 +1,47 @@
...
+	RefreshToken string `json:"refresh_token"`
+	Scope        string `json:"scope"`
+	IDToken      string `json:"id_token"`
+	TokenType    string `json:"token_type"`
+	ExpiresIn    int    `json:"expires_in"`
+	Expiry       time.Time
+}
+
+func (t *Token) SetAuthHeader(r *http.Request) {
+	r.Header.Set("Authorization", "Bearer "+t.AccessToken)
```
andreaangiolillo, line 33:
```suggestion
	r.Header.Set("Authorization", fmt.Sprintf("Bearer %s", t.AccessToken))
```

andreaangiolillo (commented):
LGTM. Nothing blocking!

andreaangiolillo (approved):

auth/oauth.go
gssbzn: > If we got an error, and the context has been canceled

https://pkg.go.dev/context

andreaangiolillo: Sorry, I didn’t explain well myself. I don’t understand in which scenarios we may get an error without the context.

gssbzn: if this scenario happens because the context is cancelled, then we don't need to check the context error if the context has _not_ been cancelled

auth/token.go
gssbzn: string concatenation via `+` is way faster (https://dev.to/pmalhaire/concatenate-strings-in-golang-a-quick-benchmark-4ahh) given how simple the string is (appended at the end) I don't think formatting here is worth it

auth/device_flow.go
```diff
@@ -0,0 +1,99 @@
...
+
+// ErrTimeout is thrown when polling the server for the granted token has timed out.
+var ErrTimeout = errors.New("authentication timed out")
+
+// PollToken polls the server until an access token is granted or denied.
+func (c Config) PollToken(ctx context.Context, code *DeviceCode) (*Token, *atlas.Response, error) {
+	timeNow := code.timeNow
+	if timeNow == nil {
+		timeNow = time.Now
+	}
```
threebee, line 64: [question for my own learning] will this ever trigger? `timeNow func() time.Time` will always return time?

auth/oauth.go
```diff
@@ -0,0 +1,197 @@
...
+	"runtime"
+	"strings"
+
+	atlas "go.mongodb.org/atlas/mongodbatlas"
+)
+
+const defaultBaseURL = atlas.CloudURL
+
+var (
+	userAgent = fmt.Sprintf("go-mongodbatlas/%s (%s;%s)", atlas.Version, runtime.GOOS, runtime.GOARCH)
```
threebee, line 34: [possible future improvement] use the same user agent across the codebase
https://github.com/mongodb/go-client-mongodb-atlas/blob/80a3a1b7eb84c9224e648a1ea30ff9d1ffd44850/mongodbatlas/mongodbatlas.go#L48

threebee (approved):
Code flow LGTM, I did not go deeper in Golang specifics due to my limited working knowledge but I did learn quite a lot from this PR. Thank you for that!

auth/token.go
andreaangiolillo: Can we really use this article as proof that this is correct? looking at the comments many developers don’t agree. For instance:

Screenshot 2022-01-25 at 12 26 52

Anyway, this is not blocking at all : ). Thanks for sharing the article

auth/device_flow.go
gssbzn: I have done the split  but I don't think consumers need to do one request as that will usually be invalid, also splitting this has a drawback that you'll create a new request object on each iteration of the loop.

auth/token.go
gssbzn: yes, if you read that comment plus is still faster than `Sprintf`

auth/device_flow.go
andreaangiolillo: > I have done the split but I don't think consumers need to do one request as that will usually be invalid
Ack.

>  you'll create a new request object on each iteration of the loop.
I am not sure why. We could have something like:

```golang
req, error := PoolTokenRequest(...) // this set up the request
// check error
return PoolTokenRetry(req)




func ... PoolTokenRetry(req) {
 for {
 ....
 resp, err2 := c.Do(ctx, req, &r)
 ....
 }
}
```

auth/device_flow.go
gssbzn: this is a reference to the function `time.Now`, line 93 then invokes the function which will return "now" at the moment of invocation, this patters is so now can be mocked by passing a `code` with a different "now"

auth/oauth.go
gssbzn: This is not related to http.Transport, this was a workaround for old versions of the `http.Client` seems like is no longer needed

auth/device_flow.go
gssbzn: not sure what your suggestion is based on your comment, can you please continue the discussion on the code as it looks now

andreaangiolillo (dismissed):

andreaangiolillo (approved):
LGTM. Thanks for answering all my questions! : )

gssbzn:
@themantissa this should have no impact on terraform but tagging you just in case

themantissa (approved):
LGTM, thank you for the heads up.

PR #323, opened 2022-10-26
task: drop support for 1.17

gssbzn:
## Description

Drops support for go1.17, this is inline with our support policy we kept it because it wasn't affecting us but it now block updating a test lib which requires 1.18+

Link to any related issue(s): 

## Type of change:


andreaangiolillo (approved):
LGTM

PR #328, opened 2022-11-17
fix: change wrong variable name on cluster update test to avoid  misu…

atileren:
## Description

client.Clusters.Update method returns Cluster object. Cluster update method test result has assigned to wrong variable name that is related with user as dbUser. I changed it to "cluster" as return object to avoid misunderstanding .


atileren:
hi @gssbzn , I've signed the contributor license agreement

gssbzn (approved):
LGTM and CLA signed

Thanks for the contribution

PR #340, opened 2023-01-09
fix: fixes ldap config to allow PATCH

### LINKED ISSUE #339: Unable to set the "authenticationEnabled" = false using the client.LDAPConfigurations.Save()
**Describe the bug**
Im not being able to set the property, "authenticationEnabled" to FALSE using the client.LDAPConfigurations.Save(), every time this property is provided as FALSE, is discarted, and keeps being TRUE. 

in the next example, im executing the save functionality with only the AuthenticationEnabled in False, then getting the ldap configurations with a curl call, and validating that the property is in true. 

image

is possible to set that property in false using an api call 
image

ive also tried to set it on TURE (original was set on false) and it works ok
 
image

 **# Summary: the property AuthenticationEnabled can be set on TRUE, but cannot be set on FALSE**

**Expected behavior**
When providing the property authenticationEnabled in false using the client, should be set in false in the atlas api. 

**Desktop (please complete the following information):**
 - OS: macOS Monterey 12.3 
 - Lib Version go.mongodb.org/atlas v0.20.0
 - Go 1.19

gssbzn:
## Description

Fixes #339 

This will be a breaking change but it's necessary to correct the reported issue

## Type of change:

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Required Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/mongodbatlas/ldap_configurations.go
+++ b/mongodbatlas/ldap_configurations.go
@@ -56,15 +56,15 @@ type LDAPConfiguration struct {
 
 // LDAP specifies an LDAP configuration for a Atlas project.
 type LDAP struct {
-	AuthenticationEnabled bool               `json:"authenticationEnabled,omitempty"` // Specifies whether user authentication with LDAP is enabled.
-	AuthorizationEnabled  bool               `json:"authorizationEnabled,omitempty"`  // The current status of the LDAP over TLS/SSL configuration.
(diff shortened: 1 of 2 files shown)


gssbzn:
FYI @Zuhairahmed since this may have impact for terraform and/or CFN

PR #344, opened 2023-01-25
CLOUDP-155262: initial version of the v2 SDK

wtrocki:
## Description

Add required files and examples in order to tests sdk end to end. 
SDK seems to be working as it should but I found issues in the versioning API that i need to fix.


wtrocki:
Wrong PR head.

PR #345, opened 2023-01-25
CLOUDP-155285: add examples to the sdk

wtrocki:
## Description

Add required files and examples in order to test sdk end to end. 
Follow up PRs will introduce documentation, sdk generator etc.

## Updates

1. I took multi-module approach to not affect current users. As result we get extra 2 root files that aren't affecting other SDK with that in place we can possibly merge v2 branch faster without affecting v1 users with extra dependencies or other issues related with bringing another version.

2. Wrapper for API in order to maintain the same level of features between v1 and v2. 
Wrapper returns client etc.

3. Latest API root. Based on the spec and how we going to handle versioning we have an api (latest)  
and will be getting an versioned sdks in form of apiv20230101 imports for future versions.


v2/test/utils/helpers.go
```diff
@@ -0,0 +1,77 @@
+package utils
```
wtrocki, line 1: While this package and its location might look odd I have much more tests coming in the follow-up PRs. Utils are reused between examples and tests.

v2/client.go
```diff
@@ -0,0 +1,56 @@
+package v2 // import "go.mongodb.org/atlas/mongodbatlas/v2"
```
wtrocki, line 1: Based on v1 client. Porting all needed features for Atlas CLI.

v2/go.mod
```diff
@@ -0,0 +1,12 @@
+module go.mongodb.org/atlas/mongodbatlas/v2
```
wtrocki, line 1: This module is transient. Removing module will not make any difference to the end user due to use of the same path. We only use it to separate dependencies between v1

v2/client.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/mongodb-forks/digest"
+	apiv2latest "go.mongodb.org/atlas/mongodbatlas/v2/api"
+)
+
+const (
+	// CloudURL is default base URL for the services.
+	CloudURL = "https://cloud.mongodb.com/"
+	// Version the version of the current API client. Should be set to the next version planned to be released.
+	// TODO use common version across entire SDK
+	Version = "0.21.0"
```
wtrocki, line 18: Have separate issue for that as it will need testing and verification of v1

v2/examples/cluster_aws.go
```diff
@@ -0,0 +1,163 @@
...
+	// See: https://www.mongodb.com/docs/atlas/app-services/authentication/api-key/
+	apiKey := os.Getenv("MDB_API_KEY")
+	apiSecret := os.Getenv("MDB_API_SECRET")
+
+	transport := digest.NewTransport(apiKey, apiSecret)
+	httpClient, err := transport.Client()
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+	sdk := client.NewClientWithUrl(httpClient, "https://cloud-dev.mongodb.com")
```
wtrocki, line 39:
```suggestion
	sdk := client.NewClientWithUrl(httpClient, "https://cloud.mongodb.com")
```

v2/test/utils/helpers.go
gssbzn, line 1: no `utils` please [About package names](https://github.com/golang/go/wiki/CodeReviewComments#package-names)

v2/test/utils/helpers.go
```diff
@@ -0,0 +1,77 @@
...
+	conn, err := net.Dial("udp", "8.8.8.8:80")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer conn.Close()
+
+	localAddr := conn.LocalAddr().(*net.UDPAddr)
+
+	return localAddr.IP.String()
+}
```
gssbzn, line 22: may I present you to our inhouse solution
https://github.com/mongodb/go-client-mongodb-atlas/blob/d8b9275b859d6e402474a2acc292670e57435932/mongodbatlas/ip_info.go#L41-L42

v2/client.go
```diff
@@ -0,0 +1,56 @@
...
+	"net/http"
+	"runtime"
+
+	"github.com/mongodb-forks/digest"
+	apiv2latest "go.mongodb.org/atlas/mongodbatlas/v2/api"
+)
+
+const (
+	// CloudURL is default base URL for the services.
+	CloudURL = "https://cloud.mongodb.com/"
```
gssbzn, line 15: shouldn't this be the default and if you plan to make use this package for tests then i would recommend to make it configurable

v2/client.go
gssbzn, line 18: let's go there
```suggestion
	Version = "0.99.0"
```

v2/client.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+*  NewSDKClientWithCredentials Helper method used to create an SDK object for provided API keys
+ */
+func NewSDKClientWithCredentials(apiKey string, apiSecret string) *apiv2latest.APIClient {
+	transport := digest.NewTransport(apiKey, apiSecret)
+	httpClient, err := transport.Client()
+	if err != nil {
+		log.Fatalf(err.Error())
```
gssbzn, line 53: is this for testing or user consumption? if for user consumption return err instead of fatal

v2/examples/cluster_aws.go
```diff
@@ -0,0 +1,163 @@
...
+/*
+* MongoDB Atlas Go SDK Example
+*
+* Example performs following actions:
+* - Finds first project in the user organization
+* - Creates new MongoDB cluster backed by AWS provider in that project
+* - Creates database user for connection
+* - Creates IP access to enable connection
+ */
+func main() {
```
gssbzn, line 27: Based on your description the goal of this is for testing, two options, make this e2e tests, see the CLI for example how it's organize https://github.com/mongodb/mongodb-atlas-cli/blob/master/test/e2e/atlas/clusters_file_test.go or consider proper [go examples](https://go.dev/blog/examples)

Moving to proper test removes the complexity of the workspace as test will be flagged as such and not included in final builds

gssbzn (commented):
some comments here happy we are moving forward with this

v2/client.go
wtrocki: URL is configurable. Changing name:
```suggestion
	DefaultCloudURL = "https://cloud.mongodb.com/"
```

v2/client.go
wtrocki: WHY?
I got information that for our sanity we need to flag or prefix version to differentiate between old and new SDK. 
That means that the client name should change and the version should be different. 

Removing TODO

v2/examples/cluster_aws.go
wtrocki: The intention was always to have it as an end-user example. 
We will compile it. Users can copy and paste the code to their repository etc.
We have separate tests that we will likely now move to Atlas CLI.

wtrocki:
Generally files aren't documented. I will create separate PR with just readme that will link with usage etc.

wtrocki:
Please ignore latest commit. It was merged based on other PR approvals

v2/test/generators/generators.go
```diff
@@ -0,0 +1,66 @@
...
+		n := num.Int64()
+		// Make sure that the number/byte/letter is inside
+		// the range of printable ASCII characters (excluding space and DEL)
+		if n > 64 && n < asciiMax {
+			result += string(rune(n))
+		}
+	}
+}
+
+func GenerateUniqueName(prefix string) (string, error) {
```
gssbzn, line 50: [avoid repetition](https://go.dev/blog/package-names#naming-package-contents) `generators.RandomIntger` vs `generators.GenerateRandomInteger`
```suggestion
func RandomInteger(min int, max int) (int, error) {
	offset, err := rand.Int(rand.Reader, big.NewInt(int64(max-min)))
	if err != nil {
		return 0, err
	}
	return min + int(offset.Int64()), err
}

func RandomASCIIString(length int) (string, error) {
	const asciiMax = 127
	result := ""
	for {
		if len(result) >= length {
			return result, nil
		}
		num, err := rand.Int(rand.Reader, big.NewInt(int64(asciiMax)))
		if err != nil {
			return "", err
		}
		n := num.Int64()
		// Make sure that the number/byte/letter is inside
		// the range of printable ASCII characters (excluding space and DEL)
		if n > 64 && n < asciiMax {
			result += string(rune(n))
		}
	}
}

func UniqueName(prefix string) (string, error) {
```

gssbzn (commented):
almost there

gssbzn (approved):
LGTM

PR #346, opened 2023-01-26
CLOUDP-155262: move required fields for source generation

wtrocki:
Target for this PR is to move all fields that we will reuse in the atlas-sdk generation. 
No file changes were made.


gssbzn (approved):
LGTM

PR #347, opened 2023-01-26
CLOUDP-155262: sdk required tools

wtrocki:
## Description

Required tools for SDK. New files only in [1b19c45](https://github.com/mongodb/go-client-mongodb-atlas/pull/347/commits/1b19c459d2d56f12c77ac410018b1b54c0ffc90b)

### DIFF
--- a/v2/tools/package-lock.json
+++ b/v2/tools/package-lock.json
@@ -1,0 +1,2092 @@
+{
+  "name": "openapi-generation-tools",
+  "version": "0.1.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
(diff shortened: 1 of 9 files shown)


gssbzn (approved):
LGTM since this is a simple move but let's add a point to discuss the expectations of javascript on this repo since so far is not something we've done

wtrocki:
> LGTM since this is a simple move but let's add a point to discuss the expectations of javascript on this repo since so far is not something we've done

+1. This is a temporary change. Before reaching other integrations we should make call if the transformer would be a permanent solution or regular part of the tooling.

PR #348, opened 2023-01-26
CLOUDP-155262: scripts and automation for source code generation

wtrocki:
## Description

Minimal scripts and automation for handling versioned SDK:
Scripts separate different concerns:

- Fetching inputs
- Processing and Generation

Follow up: CI/CD process using that scripts.

### DIFF
--- a/v2/Makefile
+++ b/v2/Makefile
@@ -1,0 +1,26 @@
+## Run examples
+
+.PHONY: example
+example:
+	go run ./examples/example_cluster_aws.go
+
+# Tools 
+.PHONY: clean
+clean:
+	rm -Rf ./api/** ./openapi/**
+
+.PHONY: fetch
+fetch:
+	./tools/scripts/fetch.sh
+
+.PHONY: generate
+generate:
+	(cd ./tools && \
+	OPENAPI_FOLDER="../openapi" SDK_ROOT="../" \
+	./scripts/generate.sh)
+
+.PHONY: validate
+validate:
+	echo "TODO"
+
+
--- a/v2/Makefile
+++ b/v2/Makefile
@@ -19,8 +19,4 @@ generate:
 	OPENAPI_FOLDER="../openapi" SDK_ROOT="../" \
 	./scripts/generate.sh)
 
-.PHONY: validate
-validate:
-	echo "TODO"
-
 
--- a/v2/tools/scripts/fetch.sh
+++ b/v2/tools/scripts/fetch.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -e
+
+## Input variables with defaults
+
+## Locked version of the versioned API
+CURRENT_REVISION=${CURRENT_REVISION:-"2023-02-01"}
+
+## OpenAPI file (latest)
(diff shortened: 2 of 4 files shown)


wtrocki:
Rebased

v2/Makefile
```diff
@@ -0,0 +1,26 @@
...
+.PHONY: generate
+generate:
+	(cd ./tools && \
+	OPENAPI_FOLDER="../openapi" SDK_ROOT="../" \
+	./scripts/generate.sh)
+
+.PHONY: validate
+validate:
+	echo "TODO"
+
```
wtrocki, line 25:
```suggestion
```

v2/tools/scripts/generate.sh
```diff
@@ -0,0 +1,26 @@
...
+OPENAPI_FILE="$OPENAPI_FOLDER/atlas-api.yaml"
+TRANSFORMED_FILE="$OPENAPI_FOLDER/atlas-api-transformed.yaml"
+
+## go package containing rolling updates for all recent versions 
+LATEST_PACKAGE=${LATEST_PACKAGE:-"api"}
+SDK_ROOT=${SDK_ROOT:-"./"}
+
+echo "Running generation pipeline"
+
+## TODO should be supported by transformer itself
```
wtrocki, line 14:
```suggestion
```

v2/tools/scripts/fetch.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
gssbzn, line 1: what the difference between this folder and https://github.com/mongodb/go-client-mongodb-atlas/tree/master/scripts

v2/tools/scripts/fetch.sh
```diff
@@ -0,0 +1,33 @@
...
+set -e
```
gssbzn, line 2: can you please make sure these bash files are passing https://github.com/koalaman/shellcheck and ok to do a follow up to add the [shellcheck gh action](https://github.com/mongodb/mongodb-atlas-cli/blob/master/.github/workflows/shellcheck.yml)?

v2/tools/scripts/fetch.sh
wtrocki: Great question. I wasn't sure if we should mix v1 and v2 tooling and scripts. If you do not see problem with moving all v2 helpers into root repo it might actually make things much cleaner as v2 will only have source code.

gssbzn: I'll allow it

Joke aside no strong opinion but like you mention grouping scripts in a common script folder I think would make it clearer

gssbzn (approved):

v2/tools/scripts/fetch.sh
wtrocki: I will go with it. Tradeoff would be that we will mix v2 and v1 files which will make it harder to clean v1 version but that might be very long time anyway.

wtrocki:
I would apply both spellcheck and script refactor in the separate PR as folder refactor introduces a lot of extra changes etc.

PR #355, opened 2023-01-30
CLOUDP-153472: Apply allOf transformation globally

AndreiMatei2206:
## Description
Divide the `allOf` transformation in two steps: expand parent and flatten child properties. Apply the `allOf` transformation to all models that qualify:
- have `oneOf`
- have `properties`


wtrocki:
I think we need to:

- Keep extensions (opt-in mechanism) still present
- Enable validation for detecting missing extensions in the openapi schema
- Make the "apply-all" step optional

wtrocki:
I think we are closely ready to bring those changes into the ./tools. 

I put some thoughts and I feel that we can archieve validation of the cases by:

- [ ] Update PR to latest changes/standards in the repository
- [x] Apply transformation out of the box for all the cases we see allOf and oneOf specifically for enums
- [ ] Update documentation
- [ ] Removing X-XGEN-Transformer annotations and move all the cases to a separate exception list.
- [ ] Introduce validator that does fail if OpenAPI is not meeting criterias that are in the exceptions list. 

## Open questions

1. I'm not sure if we need opt out mechanism - if we were doing transformation automatically and for some reason it would break some cases maybe we need to be able to opt out the same way like we do validation?

wtrocki:
@AndreiMatei2206 I have updated branch from master. Please consider running:

git commig reset --soft head~3 - where 3 is number of your commits and pushing your changes again.

api/v1alpha/README.md
```diff
@@ -321,6 +321,7 @@ Class | Method | HTTP request | Description
  - [AWSAutoScaling](docs/AWSAutoScaling.md)
  - [AWSCloudProviderContainer](docs/AWSCloudProviderContainer.md)
  - [AWSCloudProviderContainerAllOf](docs/AWSCloudProviderContainerAllOf.md)
+ - [AWSCloudProviderContainerAllOf1](docs/AWSCloudProviderContainerAllOf1.md)
```
wtrocki, line 324: I remember we wanted to run a transformation to remove AWSCloudProviderContainerAllOf. Do we know why we have a new object introduced?

api/v1alpha/api/openapi.yaml
```diff
@@ -61296,8 +61284,8 @@ components:
       type: object
     AWSCloudProviderContainer:
       allOf:
-      - $ref: '#/components/schemas/CloudProviderContainer'
       - $ref: '#/components/schemas/AWSCloudProviderContainer_allOf'
+      - $ref: '#/components/schemas/AWSCloudProviderContainer_allOf_1'
```
wtrocki, line 61288: We do have very strange allOf of the allOf object introduced.

tools/scripts/transform.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
wtrocki, line 1: Should we have transformation as a separate step (from generation)?
Sounds like a good idea although we need to update GitHub action with an extra step (I can help with that in separate PR)

tools/transformer/src/transformations.js
```diff
@@ -133,13 +129,18 @@ function transformAllOf(objectName, parentObject, api) {
...
       // Flatten child allOf
-      flattenAllOfObject(child, api);
     }
   }
+
+  // Remove invalid fields
+  parentObject.properties = null;
+  delete parentObject.properties;
+
+  parentObject.required = null;
```
wtrocki, line 142: that would be affecting client-side validation. Is that desired?

wtrocki: I see that it was moved, just wondering

tools/transformer/src/transform.js
```diff
@@ -21,9 +21,9 @@ if (apiFileLocation) {
   throw new Error("Missing location. Please set OPENAPI_FILE env variable");
 }
 
+doc = applyModelNameTransformations(doc, "ApiAtlas", "View");
+
 doc = applyOneOfTransformations(doc);
 doc = applyAllOfTransformations(doc);
```
wtrocki, line 27: applyAllOfTransformations  - should happen first before  applyOneOfTransformations

tools/transformer/src/transformations.js
AndreiMatei2206: I was trying to figure out a bug, and I thought that I was deleting by mistake some properties and those would be garbage collected. I will revert those changes

tools/transformer/src/transform.js
AndreiMatei2206: Yup missed that - will update

wtrocki:
Reviewed. I see a huge positive impact of this change. You have identified and fixed number of extra (undefined) allOf transformations that introduced missing fields. We need to understand why we are getting `allOf of allOf` objects in the output schema and see if we can avoid it. Time issues are likely result of the wrapper of the wrapper.

tools/transformer/src/transformations.js
wtrocki: Actually you are right. If properties are required but we remove them openapi schema will no longer be valid.

api/v1alpha/api/openapi.yaml
AndreiMatei2206: During debugging I deleted the function that was flattening the object - re-added it now and this should be gone

tools/scripts/transform.sh
AndreiMatei2206: Sure sounds good! I can try to add it on this PR

api/v1alpha/api/openapi.yaml
wtrocki: That was fast fix for someting that sounded like long investigation.

Ignore my comment in: https://github.com/mongodb/go-client-mongodb-atlas/pull/355#issuecomment-1430160166

api/v1alpha/api/openapi.yaml
```diff
@@ -61866,14 +62148,6 @@ components:
           - ASIA_SOUTHEAST_2
           type: string
           x-xgen-go-transform: merge-oneOf
-        analyticsAutoScaling:
```
wtrocki, line 61869: Those are child fields (AWS region config) that were removed. 
Do we know why? Is that ok?

api/v1alpha/model_aws_region_config_all_of.go
```diff
@@ -0,0 +1,223 @@
...
+// Code generated by OpenAPI Generator (https://openapi-generator.tech); DO NOT EDIT.
+
+package v1alpha
+
+import (
+	"encoding/json"
+)
+
+// AWSRegionConfigAllOf struct for AWSRegionConfigAllOf
+type AWSRegionConfigAllOf struct {
```
wtrocki, line 18: This file should not be present as it doesn't have representation in model.
We need to add instruction on how to regenerate model locally (Update readme to run make command that will remove SDK.

wtrocki:
Looks good. Will approve after some minor tweaks and to some extra verification and testing.

tools/transformer/src/transform.js
AndreiMatei2206: Actually on this - there are some cases where we are doing oneOf transformations to models that would qualify for the allOf transformation, i.e. `ApiAtlasCloudProviderAccessRoleView` (but more importantly), `ApiAtlasDedicatedHardwareSpecView`. So I think it would be best to keep the transformations in this order.

api/v1alpha/model_aws_region_config_all_of.go
AndreiMatei2206: Fixed

tools/transformer/src/transformations.js
```diff
@@ -91,49 +115,29 @@ function transformAllOf(objectName, parentObject, api) {
     );
   }
 
-  const parentProperties = parentObject.properties;
+  const expandedParent = filterObjectProperties(parentObject, (key, _v) => {
+    return !(key === 'oneOf' || key === 'discriminator');
```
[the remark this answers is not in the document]
AndreiMatei2206: Here I am getting the properties of the parent, excluding the oneOf and discriminator if present, i.e. I am only interested in parent.properties and parent.required (and maybe other extensions if we want to merge them in the future). It acts as a exclusion list for what properties of the parent to keep.

For the "discriminator should always have an equivalent oneOf" will be done as part of the investigation of removing oneOf [CLOUDP-156197](https://jira.mongodb.org/browse/CLOUDP-156197)

tools/transformer/src/transform.js
wtrocki: Sounds good. We can handle oneOf propery transformation in separate PR. Curious what sdk would look like we we also apply that across the board

tools/transformer/src/utils.js
```diff
@@ -88,28 +106,119 @@ function detectDuplicates(objArray) {
...
+  delete obj["$ref"];
+
+  Object.keys(dereferencedObj).forEach((key) => {
+    obj[key] = dereferencedObj[key];
+  });
+
+  return obj;
+}
+
+function flattenAllOfObject(obj, apiObject) {
```
wtrocki, line 124: Btw. This is no longer utils. It is actual logic of AllOf transformation. 
Unless it is resused it should be probably outside utils.

wtrocki (approved):

PR #357, opened 2023-01-31
CLOUDP-155253: workaround accept header negotiation in versioned api

wtrocki:
## Motivation

Workaround problem with versioning API when accept header contains application/json type.
Please ignore *.go files to simplify review. After that change all examples work end to end.

## Solution

Use custom generator template with minimal change required to enable requests for versioning APIs.
This solution is temporary. I have made change only in single file rather than larger refactor to make template simpler to remove and update in future.

## Rejected solutions

1. Transformation of the openapi file by reusing responses with versioned content type.
Generated large openapi file (3 times more than we have right now).

2. Transformation using specific versions.
Generating responses for specific versions and reusing them. 
It required transformer to be version aware

3. Removing content type from responses 
The client lost all error parsing. JSON data weren't available to users.


v2/tools/transformer/src/transformations.js
```diff
@@ -164,25 +164,8 @@ function transformOneOf(parentObject, api) {
...
 }
 
-// Temporary transformation for versioned api
-function applyResponseTransformation(doc){
-  const hasResponses = doc.components && doc.components.responses 
-  if (hasResponses) {
-    for (const responseKey of Object.keys(doc.components.responses)) {
-      response = doc.components.responses[responseKey];
-      if (response.content && response.content["application/json"]) {
-        response.content["application/vnd.atlas.unversioned+json"] =
```
wtrocki: Removed transformation. 
This transformation did not work properly as the versioned API only supports singular content type. We could only use custom template.

v2/tools/config/go-templates/client.mustache
```diff
@@ -0,0 +1,635 @@
...
+}
+
+// selectHeaderAccept join all accept types and return
+func selectHeaderAccept(accepts []string) string {
+	if len(accepts) == 0 {
+		return ""
+	}
+	{{! X-XGEN-MODIFIED }}
+	{{! Versioned API adds limitation for clients to send only one content type. No support for content negotiation}}
+	return accepts[0]
```
AndreiMatei2206, line 105: Are we sure that this is always going to return the "application/json" and not the ""application/vnd.atlas.2023-01-01+json"?

AndreiMatei2206 (approved):
LGTM-ing, just a small  non-blocking concern

PR #362, opened 2023-02-02
CLOUDP-156527: add regression tests

wtrocki:
I am adding regression tests to detect possible breaking changes in the happy path. 
Those tests would be only temporary as we would have breaking changes detection done by versioned API

### DIFF
--- a/v2/test/regression/multi_cloud_clusters.go
+++ b/v2/test/regression/multi_cloud_clusters.go
@@ -1,0 +1,151 @@
+package regression
+
+import (
+	mongodbatlas "go.mongodb.org/atlas/mongodbatlas/v2/api"
+)
+
+// Backwards compatibility, regression tests for the multi cloud clusters
+func TestMultiCloudClusters_ModelRegression() {
+	_true := true
+	_string := ""
+	_int := int32(0)
+	labelView := mongodbatlas.NDSLabel{
+		Key:   &_string,
+		Value: &_string,
+	}
+
+	instanceSize := mongodbatlas.INSTANCESIZE_M10
+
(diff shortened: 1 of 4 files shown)


v2/test/regression/multi_cloud_clusters.go
```diff
@@ -0,0 +1,151 @@
...
+		Key:   &_string,
+		Value: &_string,
+	}
+
+	instanceSize := mongodbatlas.INSTANCESIZE_M10
+
+	autoScale := &mongodbatlas.AutoScalingV15{
+		Compute: &mongodbatlas.ComputeAutoScalingV15{
+			Enabled: &_true,
+			// TODO missing oneOf transformation
```
gssbzn, line 22: can you link a CLOUDP?

v2/test/regression/multi_cloud_clusters.go
```diff
@@ -0,0 +1,151 @@
...
+
+import (
+	mongodbatlas "go.mongodb.org/atlas/mongodbatlas/v2/api"
+)
+
+// Backwards compatibility, regression tests for the multi cloud clusters
+func TestMultiCloudClusters_ModelRegression() {
+	_true := true
+	_string := ""
+	_int := int32(0)
```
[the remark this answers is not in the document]
gssbzn: or with generics now you could implement the function yourself 
```go
func Pointer[T any](x T) *T {
	return &x
}
```

gssbzn (approved):
some small comments not blocking

v2/test/regression/multi_cloud_clusters.go
wtrocki: We addressed that already. Good catch

PR #363, opened 2023-02-02
CLOUDP-155226: include versioned API version in requests

wtrocki:
## What

Use env var provided version when running generator job.

## Why

To fully automate generation we need to supply all input arguments from CI/CD job.
The current version matches the proposal in the specification and it is likely to change.

### DIFF
--- a/v2/api/atlas_client.go
+++ b/v2/api/atlas_client.go
@@ -12,7 +12,7 @@ const (
 	// CloudURL is default base URL for the services.
 	DefaultCloudURL = "https://cloud.mongodb.com/"
 	// Version the version of the current API client. Should be set to the next version planned to be released.
-	Version    = "1.230201.0-dev1"
+	Version    = "1.20230201.0-dev1"
 	ClientName = "go-mongodbatlas-versioned"
 )
 
--- a/v2/api/atlas_client.go
+++ b/v2/api/atlas_client.go
@@ -12,7 +12,7 @@ const (
 	// CloudURL is default base URL for the services.
 	DefaultCloudURL = "https://cloud.mongodb.com/"
 	// Version the version of the current API client. Should be set to the next version planned to be released.
-	Version    = "2.0.0-dev1"
+	Version    = "1.230201.0-dev1"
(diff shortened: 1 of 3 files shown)


v2/tools/scripts/generate.sh
```diff
@@ -4,6 +4,8 @@ set -e
 OPENAPI_FOLDER=${OPENAPI_FOLDER:-"./openapi"}
 OPENAPI_FILE="$OPENAPI_FOLDER/atlas-api.yaml"
 TRANSFORMED_FILE="$OPENAPI_FOLDER/atlas-api-transformed.yaml"
+## Version of the versioned API
+CLIENT_VERSION=${CLIENT_VERSION:-"1.230201.0-dev1"}
```
gssbzn, line 8: let's keep it YYYYMMDD
```suggestion
CLIENT_VERSION=${CLIENT_VERSION:-"1.20230201.0-dev1"}
```

wtrocki: Agreed. Updated spec as well.

v2/api/atlas_client.go
```diff
@@ -12,7 +12,7 @@ const (
 	// CloudURL is default base URL for the services.
 	DefaultCloudURL = "https://cloud.mongodb.com/"
 	// Version the version of the current API client. Should be set to the next version planned to be released.
-	Version    = "2.0.0-dev1"
+	Version    = "1.230201.0-dev1"
```
gssbzn, line 15: would this be autoupdated? good to do as a follow up tho

gssbzn (approved):
LGTM

v2/api/atlas_client.go
wtrocki: I'm going to set env var in the generator job and we should see it. Left it intentionally to have the effect in the PR.

wtrocki: Actually merging PR with wrong value seems wrong. Will regenerate it now

wtrocki:
```suggestion
	Version    = "1.20230201.0-dev1"
```

PR #365, opened 2023-02-02
CLOUDP-156539: notifications only for v0

wtrocki:
## Description

Reducing the amount of noise happening in the new SDK. 
We need apix to still receive useful notifications to manually written client.
More context in the jira.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,3 @@
 # Maintained by the API experience team
-*       @mongodb/apix
+mongodbatlas/*       @mongodb/apix
+auth/*               @mongodb/apix


.github/CODEOWNERS
```diff
@@ -1,2 +1,3 @@
 # Maintained by the API experience team
-*       @mongodb/apix
+mongodbatlas/*       @mongodb/apix
```
wtrocki, line 2: the new version of the client will use a different notification mechanism. 
This will keep the current notification mechanism for v0 and stop unneeded noise for changes driven by generated client.

gssbzn: this leaves out changes to the go.mod/go.sum files, so we won't get notified on dependabot changes, same with the gh actions folders

gssbzn: I was reading the docs for codeowners and seems like the not(!) operator is not supported :(

wtrocki:
Closing. I think we need to remove code owners in the long run and replace them with better integration. 
No need for any action now.

PR #372, opened 2023-02-09
CLOUDP-156178: CI/CD setup for tools folder

wtrocki:
Using default prettier profile and running jest unit tests.
[fbc5121](https://github.com/mongodb/go-client-mongodb-atlas/pull/372/commits/fbc5121ebdac12bf3a0fd9217f234d54f03af595) doesn't require review as it is only introducing formatting changes.


.github/workflows/tools.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - "tools/**"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
gssbzn, line 14: can we use the latest LTS (18) instead of 16 which is going EOL later this year?

tools/package.json
```diff
@@ -8,9 +8,15 @@
     "yaml": "2.2.1"
   },
   "devDependencies": {
-    "@openapitools/openapi-generator-cli": "2.5.2"
```
gssbzn, line 11: can you please include in this file the node version and npm versions?

.github/workflows/tools.yaml
wtrocki: Updating. I did used 16 as this is what we are using right now.

.github/workflows/tools.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - "tools/**"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
```
gssbzn, line 14:
```suggestion
          node-version: 18
          cache: 'npm'
```

gssbzn (commented):
let's enable cache for the dependencies

.github/workflows/tools.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - working-directory: ./tools
+        run: |
+          npm install
+          npm ci
```
gssbzn, line 18: are you sure this is running your `ci` script instead of the `npm-ci - Clean install a project` command?
I think for your script you may need to do 
```suggestion
          npm run ci
```

wtrocki:
Good catch about cache. I assumed it is enabled by default. Build seems to be working now

gssbzn (approved):
LGTM awesome work here

PR #383, opened 2023-02-20
CLOUDP-162772: Apply oneOf transformation globally for enums

AndreiMatei2206:
## Description

Apply oneOf transformation to objects that:
- have oneOf property
- all the objects in the oneOf are enums


wtrocki:
Do we also need to update readme.md file?

api/v1alpha/api/openapi.yaml
```diff
@@ -49004,7 +49035,49 @@ components:
           readOnly: true
           type: string
         eventTypeName:
-          $ref: '#/components/schemas/DefaultAlertViewForNdsGroup_eventTypeName'
+          description: Incident that triggered this alert.
```
wtrocki, line 49038: [Unrelated to pr]
Coming from the point of doing generally nonharmful transformations would we consider adding prefix to each enum. This way any overwrite will be persistent. 

It doesn't affect SDK in any way as it does gets ignored but it will affect docs and other tools reading that schema.

WDYT about this? Is that jira worthy?

api/v1alpha/api/openapi.yaml
```diff
@@ -49220,7 +49434,66 @@ components:
           readOnly: true
           type: string
         eventTypeName:
-          $ref: '#/components/schemas/DefaultEventViewForOrg_eventTypeName'
+          description: Unique identifier of event type.
```
wtrocki, line 49437: I'm amazed how many of those are single reference meaning that oneOf on those were completely pointless.

tools/transformer/src/transformations.js
```diff
@@ -148,61 +139,74 @@ function transformAllOf(objectPath, api) {
...
-    if (oneOfObject.properties && oneOfObject.enum) {
-      const errorData = JSON.stringify(oneOfObject);
-      throw new Error(
-        `Object cannot contain both enum and properties: ${errorData}`
-      );
-    }
+  if (isEnum) {
+    transformOneOfEnum(parentObject, api);
+  } else {
+    transformOneOfObject(parentObject, api);
```
wtrocki, line 160:
```suggestion
    transformOneOfProperties(parentObject, api);
```

wtrocki: [IMHO we will never be able to run properties transformation out of the box as it will literally remove all oneOf from the code. Do you think we can experiment with the transformation to see how our SDKs will look like in that sense? How many duplicates we will get? Would UX will be affected etc.?

tools/transformer/src/transformations.js
```diff
@@ -148,61 +139,74 @@ function transformAllOf(objectPath, api) {
...
+  }
+
+  parentObject.enum = [...parentObject.enum];
+
+  // Remove invalid fields
+  delete parentObject.discriminator;
+  delete parentObject.oneOf;
+}
+
+function transformOneOfObject(parentObject, api) {
```
wtrocki, line 188: I will add some comment to the top level functions

tools/transformer/src/transformations.js
```diff
@@ -214,6 +218,23 @@ function isAllOfTransformable(obj) {
   return obj.properties && obj.oneOf;
 }
 
+function isOneOfTransformable(obj, api) {
```
wtrocki, line 221:
```suggestion
function canApplyOneOfTransformationToObject(obj, api) {
```

wtrocki (approved):
LGTM.

It looks like we found some missing alert view enums that would have been definitely causing problems when using that API. We need to update docs

tools/transformer/README.md
```diff
@@ -22,8 +22,10 @@ OPENAPI_FILE=openapi.yaml node ./transform.js
 ## Implemented transformations
 
 1. oneOf transformation
+Appliet to all objects that meet either of the following criteria:
```
wtrocki, line 25:
```suggestion
Applied to all objects that meet either of the following criteria:
```

tools/transformer/README.md
```diff
@@ -22,8 +22,10 @@ OPENAPI_FILE=openapi.yaml node ./transform.js
...
 
-> "x-xgen-go-transform": "oneOfMerge"
+> Have extension "x-xgen-go-transform": "oneOfMerge"
+> Have "oneOf" field and all the objects referenced are enums
```
wtrocki, line 28:
```suggestion
> Have "oneOf" field with all referenced schemas having an enum type
```

wtrocki (approved):

PR #391, opened 2023-02-23
CLOUDP-163498: Using SDK filter for OpenAPI

wtrocki:


wtrocki:
Removing filter completely. We will remove those in transformation instead.

gssbzn (approved):
LGTM

PR #392, opened 2023-02-24
CLOUDP-163737: go templates

wtrocki:
Add patterns for working with openapi templates.
All templates are now kept in the repo so we can see diff when we do update openapi generator

We also introduce standard for the modification of existing patterns (although it should be generally discouraged and seen as last resort solution)


gssbzn (approved):
LGTM

PR #396, opened 2023-03-01
task: drop pointy

gssbzn:
## Description

Pointy is no longer necessary with generics we can have our own implementation

Pointy has also been a source of issues with snyk which would flag its dependency on yaml.v2 every now and then


wtrocki (approved):
LGMT.
I'm going to rebase sdkv2 on top of those changes as well.

PR #400, opened 2023-03-02
Update shellcheck job to run only when sh files are changed

wtrocki:
## Description

WHY: Minimizing the number of runs of jobs for scheduled PRs.

### DIFF
--- a/tools/scripts/fetch.sh
+++ b/tools/scripts/fetch.sh
@@ -2,13 +2,13 @@
 set -o errexit
 set -o nounset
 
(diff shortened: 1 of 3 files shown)


tools/scripts/fetch.sh
```diff
@@ -2,13 +2,13 @@
 set -o errexit
 set -o nounset
 
-#######################################
```
wtrocki, line 5: Changes used to test

gssbzn (approved):

PR #401, opened 2023-03-03
CLOUDP-165150: Reduce logging for CI/CD and validation purposes.

wtrocki:
## Description

Required fixes in the transformer before we start adding linting and flagging invalid cases:

- Minimize logging for end users (do not run debug or trace commands) so output of the transformer can be used to figure out what is wrong with the openapi. Warn will be default. 
- Add abstraction for file storage (so we can introduce more ways we read and store openapi file - it will enable us to use transformer as validator case.)

## Verification

Done `make transform_openapi`
We now only see ERRORS and WARNINGS. Some of the warnings will be downgraded to info/

### DIFF
--- a/tools/transformer/__tests__/transformations.test.js
+++ b/tools/transformer/__tests__/transformations.test.js
@@ -39,17 +39,6 @@ test("Transform AllOf model", () => {
   ).toMatchInlineSnapshot(cases.ChildAllOf);
 });
 
-test("Fail Transform AllOf with duplicates", () => {
-  api.components.schemas.ApiAtlasAWSRegionConfigView.properties = {
-    ...api.components.schemas.ApiAtlasAWSRegionConfigView.properties,
-    ...api.components.schemas.ApiAtlasRegionConfigView.propertries,
-  };
-  global.console.warn = jestGlobal.fn();
-  transformAllOf(".components.schemas.ApiAtlasRegionConfigView", api);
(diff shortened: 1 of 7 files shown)


tools/transformer/__tests__/transformations.test.js
```diff
@@ -39,17 +39,6 @@ test("Transform AllOf model", () => {
   ).toMatchInlineSnapshot(cases.ChildAllOf);
 });
 
-test("Fail Transform AllOf with duplicates", () => {
```
wtrocki, line 42: We no longer need to present that to the user. This is happening for each transformation we have.

tools/transformer/src/transform.js
```diff
@@ -1,29 +1,21 @@
...
 } = require("./transformations");
-const { parse, stringify } = require("yaml");
+const { getAPI, saveAPI } = require("./apifile");
 
-let apiFileLocation = process.argv.slice(2);
-console.error(apiFileLocation);
-if (!apiFileLocation && !apiFileLocation[0]) {
-  throw new Error("Missing positional argument for openapi file");
-}
+const log = require("simple-node-logger").createSimpleLogger();
```
wtrocki, line 8: This one guarantees compatibility with the console format so we can reuse the same interface and swap to anything else we need.

AndreiMatei2206 (approved):
LGTM! Nice optimization

PR #402, opened 2023-03-03
CLOUDP-165150: refactor complex transformation code

wtrocki:
## Description

Refactor for transformer code to enable separation of transformer functions into separate files.
This is a step for introduction of validation and transforms phases for each transformation (enforcing transformation interface).

NOTE: content is exactly the same, algorithms were not changed. 
I just moved implementations to different files to separate very long and overly complex single-file implementations.


gssbzn (approved):

tools/transformer/src/wrappers/apifile.js
```diff
@@ -3,6 +3,7 @@ const { readFileSync, writeFileSync } = require("fs");
 
 /**
  * Read and parse API file as json
+ *
```
AndreiMatei2206, line 6: [q] I'm curious to know what the thought process behind naming this "wrappers" - this makes me things of wrapper functions from Python, but maybe I have a wrong understanding of what wrapper is.

AndreiMatei2206 (approved):
Overall LGTM, just have a small non-blocking question

tools/transformer/src/wrappers/apifile.js
wtrocki: Nice! I will move it to the engine to not confuse anyone.

PR #403, opened 2023-03-03
CLOUDP-165150: discriminator validation. Adding missing oneOf in transformation engine

wtrocki:
## Description

Adding discriminator transformation that performs detection of invalid cases we want to avoid in the OpenAPI. 
Ignore the second commit - that is only regenerating SDK.


tools/transformer/src/transformations/discriminator.js
```diff
@@ -0,0 +1,80 @@
...
+  const parentObject = getObjectFromYamlPath(objectPath, api);
+  const parentName = getNameFromYamlPath(objectPath);
+
+  if (!isTransformable(parentObject)) {
+    throw new Error(`Invalid object for AllOf Transformation: ${parentName}`);
+  }
+
+  if (!parentObject.oneOf && parentObject.discriminator) {
+    if (!parentObject.discriminator.mapping) {
+      throw new Error(`OpenAPI object schema contains discriminator but missing oneOf or discriminator mapping.
```
wtrocki, line 41: This is introducing a form of validation. We should not have this use case in openapi as we would not be able to figure out how to build oneOf. I did ensure that this is the case and introduced this check.

tools/transformer/src/transformations/discriminator.js_Ignore
```diff
@@ -1,274 +0,0 @@
-const {
```
wtrocki, line 1: Pushed this file by mistake. Removing that now.

tools/transformer/src/transform.js
```diff
@@ -13,6 +14,7 @@ log.setLevel(process.env.XGEN_LOGGING_LEVEL || "warn");
 
 let { doc, apiFileLocation } = getAPI(process.argv.slice(2));
 
+doc = applyDiscriminatorTransformations(doc);
```
wtrocki, line 17: Order of that operation is important. We need to run this before oneOf and allOf transformation as those will be picking up oneOfs introduced by discriminatorTransformation :)

tools/config/go-templates/model_simple.mustache
```diff
@@ -1,454 +0,0 @@
-// checks if the {{classname}} type satisfies the MappedNullable interface at compile time
```
wtrocki, line 1: The actual impact of this change. We no longer have issues with missing properties as oneOfs are everywhere and we validating them. I will restore actual template file in separate PR (we need file from the repo)

api/v1alpha/README.md
```diff
@@ -472,11 +472,8 @@ Class | Method | HTTP request | Description
  - [CloudProviderAccess](docs/CloudProviderAccess.md)
  - [CloudProviderAccessAWSIAMRole](docs/CloudProviderAccessAWSIAMRole.md)
  - [CloudProviderAccessDataLakeFeatureUsage](docs/CloudProviderAccessDataLakeFeatureUsage.md)
- - [CloudProviderAccessDataLakeFeatureUsageAllOf](docs/CloudProviderAccessDataLakeFeatureUsageAllOf.md)
```
wtrocki, line 475: CC @AndreiMatei2206 I believe that this was something you were looking for fixing. 
After adding oneOf to every discriminator - the discriminator link is no longer driving relationships. That means a much simpler result model and we do not have duplicate templates. 
This also resolves problems with os.Time

tools/transformer/src/transformations/discriminator.js
```diff
@@ -0,0 +1,80 @@
...
+      `Setting oneOf based on discriminator for allOf transformation in ${parentName}`
+    );
+    // Ignore objects that point to themselves
+    oneOfReferences = Object.values(parentObject.discriminator.mapping);
+    // Remove duplicates in referneces
+    oneOfReferences = [...new Set(oneOfReferences)];
+
+    for (referenceObj of oneOfReferences) {
+      if (getObjectNameFromReferenceString(referenceObj) == parentName) {
+        // FIXME circular reference found. We can remove that once we fix api
```
wtrocki, line 54: This entire section can be removed once we have backend linting in place.

matt-condon (approved):

AndreiMatei2206 (approved):
LGTM

wtrocki:
Thank you!

wtrocki:
FYI. This change brings an lot of SDK updates. Make sure to use `go get ..` to fetch latest from branch

PR #404, opened 2023-03-09
CLOUDP-161499: Fix names of the openapi models by transformation

wtrocki:
Only first 2 commits matter:
1: Changes error to warning
2: Reverts change but also introduces ignore list for the names we are likely to change over the time.

Rest of the changes are regenerated SDK.


tools/transformer/src/name.ignore.json
```diff
@@ -0,0 +1,13 @@
+{
```
wtrocki, line 1: This is only temporary. We have CLOUDP-165932 to remove those models from ignore list.

wtrocki:
Lint job timed out. It is passing locally.

api/v1alpha/README.md
```diff
@@ -675,97 +641,116 @@ Class | Method | HTTP request | Description
...
- - [PaginatedPrivateLinkConnectionView](docs/PaginatedPrivateLinkConnectionView.md)
- - [PaginatedRestoreJobView](docs/PaginatedRestoreJobView.md)
- - [PaginatedServerlessInstanceDescriptionView](docs/PaginatedServerlessInstanceDescriptionView.md)
- - [PaginatedSnapshotView](docs/PaginatedSnapshotView.md)
- - [PaginatedTeamRoleView](docs/PaginatedTeamRoleView.md)
- - [PaginatedTeamView](docs/PaginatedTeamView.md)
- - [PaginatedTenantRestoreView](docs/PaginatedTenantRestoreView.md)
- - [PaginatedTenantSnapshotView](docs/PaginatedTenantSnapshotView.md)
- - [PaginatedUserCertView](docs/PaginatedUserCertView.md)
+ - [PaginatedLegacyCluster](docs/PaginatedLegacyCluster.md)
```
wtrocki, line 710: That one seems like leftover:
I think the pattern should be ApiAtlasPaginated so we need to fix those names in OpenAPI instead of transformation

matt-condon (approved):
LGTM!
One question, does this mean the "ignored" models will be unavailable in the v2 client until the follow up change is made?

wtrocki:
Ignored models aren't renamed. Ignore the list that exists for us to still monitor and fail for newly introduced models. 
None of the ignored models are actually used in the SDK. Those are openapi technical debts that we going to improve over time. My goal here is to make SDK models renamed asap to make sure that names are final and CLI will never need to rename any models from now. 

Ignored models are mostly enum values that we ignore in generation process.

wtrocki:
Thank you for review :)

tools/transformer/src/transform.js
```diff
@@ -17,6 +18,10 @@ let { doc, apiFileLocation } = getAPI(process.argv.slice(2));
 doc = applyDiscriminatorTransformations(doc);
 doc = applyOneOfTransformations(doc);
 doc = applyAllOfTransformations(doc);
-doc = applyModelNameTransformations(doc, "ApiAtlas", "View");
+
+doc = applyModelNameTransformations(doc, "ApiAtlas", "View", ignoredModelNames);
```
andmatei, line 22: I think the order of this operation matters, so if we have a model `ApiAtlasApi{name}View` this sequence will transform it into `{name}` which I don't know it's the behaviour we want.

We could first remove the prefix "Api", and then from the resulting spec remove the prefix "Atlas".

```suggestion
doc = applyModelNameTransformations(doc, "Api", "", ignoredModelNames);
doc = applyModelNameTransformations(doc, "Atlas", "", ignoredModelNames);
doc = applyModelNameTransformations(doc, "", "View", ignoredModelNames);
```

tools/transformer/src/transform.js
```diff
@@ -17,6 +18,10 @@ let { doc, apiFileLocation } = getAPI(process.argv.slice(2));
...
+doc = applyModelNameTransformations(doc, "Api", "View", ignoredModelNames);
+doc = applyModelNameTransformations(doc, "", "View", ignoredModelNames);
```
andmatei, line 24: Should we deal in any way with manual views?

andmatei (commented):
A few questions, once these are clarified it's good to go.

tools/transformer/src/transform.js
wtrocki: It looks like the results output is the same. Will create follow-up PR.

PR #409, opened 2023-03-13
CLOUDP-155910: add ability to generate client from api preview

wtrocki:
## Description

API preview url might change a lot so decided to use the secret to expose it. 
I have tried to reuse the existing workflow but with a number of changes, it seems like a copy of the production one is much easier.

To reduce noise workflow is run on demand but it can be changed to a cron job as well. 
Secret created:
Screenshot 2023-03-13 at 16 10 55

### DIFF
--- a/.github/workflows/autoupdate-dev.yaml
+++ b/.github/workflows/autoupdate-dev.yaml
@@ -1,0 +1,40 @@
+name: Generate Client
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


wtrocki:
See also: https://github.com/mongodb/go-client-mongodb-atlas/pull/411

gssbzn (approved):

PR #410, opened 2023-03-13
CLOUDP-166133: 

wtrocki:
## Description

This PR changes folder structure for the v2 client:

```
api/examples -> examples
api/tests -> tests
api/README.md ->> README.md
api/v1client -> mongodbatlasv2/latest
```

Changes are based on the feedback from the CLI initiative and are also a direct result of us making a proper call for the naming. Please ignore last commit. It contains only 580 renamed files.


README.md
```diff
@@ -6,8 +6,53 @@ A Go HTTP client for the [MongoDB Atlas API](https://docs.atlas.mongodb.com/api/
...
+### Adding Dependency
+
+```
+go install go.mongodb.org/atlas
+```
+
+### Using in the code
+
+```go
+mongodbatlas import "go.mongodb.org/atlas/mongodbatlasv2/latest"
```
fmenezes, line 24: why latest? are we expecting to keep every version on master?

tools/Makefile
```diff
@@ -1,12 +1,12 @@
...
 export OPENAPI_FILE_NAME=atlas-api.yaml
-export SDK_FOLDER=../api/v1alpha
+export SDK_FOLDER=../mongodbatlasv2/latest
 
 .DEFAULT_GOAL := generate_client
 
 .PHONY: clean_client
 clean_client:
-	rm -Rf ../api/v1alpha
+	rm -Rf ../mongodbatlasv2/latest
```
fmenezes, line 9:
```suggestion
	rm -Rf $(SDK_FOLDER)
```

README.md
wtrocki: That reflects our current approach for the versioned API:

- mongodbatlasv2 (name of the client - new version of it)
- latest (metaname for the version of the versioned API we are using - currently 20230201)

If at some point we decide to host multiple versioned versions or name versions explicitly we would be able to do so.
Having just `go.mongodb.org/atlas/mongodbatlasv2` will hardcode generated client to that path and close our way to apply versioned api versions

tools/Makefile
wtrocki: I would prefer to not apply this change as relying on env var specifically introduces some risks (which happened to some upstream projects removing users hard drive)

README.md
andreaangiolillo: > If at some point we decide to host multiple versioned versions or name versions explicitly we would be able to do so.
> Having just go.mongodb.org/atlas/mongodbatlasv2 will hardcode generated client to that path and close our way to apply versioned api versions

what will block us from creating a new directory `latest` in the future if we will decide to have multiple versions? I don't understand why we cannot have only the `mongodbatlasv2` folder now and create a `latest` one In the future if needed. I agree with @fmenezes that `latest` is confusing for people who are not familiar with our versioning strategy.

wtrocki: Ok. So basically this would be premature optimization. 
Changing to: 
go.mongodb.org/atlas/mongodbatlasv2

wtrocki: Done in: [4b224e9](https://github.com/mongodb/go-client-mongodb-atlas/pull/410/commits/4b224e906fc4a9e000ca17ccfb4de5b2104eccaf)

.gitignore
```diff
@@ -37,3 +37,5 @@ node_modules
 # Output of the go coverage tool
 *.out
 example_cluster_aws
+mongodbatlasv2/api
+mongodbatlasv2/.openapi-generator
```
andreaangiolillo, line 41: Is there any reason why we go for `mongodbatlasv2` instead of `mongodbatlas`? Isn't the release version tag enough to differentiate v1 from v2? I would expect customers to import the SDK with `go.mongodb.org/atlas v2.0.0 `, they should not need to know that this is "v2" of our API, they just want to use our latest APIs

tools/Makefile
fmenezes: note that this is not an env var per se it is a makefile var, I'm ok either way though

.gitignore
wtrocki: mongodbatlas exist and it is going to be supported for months (if not years) with mongodbatlasv2 being completely different project (not just new version with couple breaking changes - it is all new - new api, clients etc.)
Lets assume implement two separate 2 release lines 0.x.0 and v2 (skipping v1?) 

Some personal worries?
- We going to lose ability to migrate projects step by step (killing incentive to update)
- Models with the same name going to produce some odd behaviour andnerrors (rather import errors)
- We going to have v0.x.x and v2.0.0 releases happening simultanously 

Benefits?

- We do not need to choose new name
- Import size is smaller.
- ?

My take is to relase both (possibly under single name and version) and at some point deprecate v1 client and subsequently remove it. Versioning can stay 0.x.0 or go into v50.x.x - personally do not care as long we have both clients available under single package for migrators.

andreaangiolillo: > We going to lose ability to migrate projects step by step (killing incentive to update)

I thought that `go.mod` supports having multiple versions of the same module but a quick search online showed that is not possible indeed 😞 .  Thanks for the discussion and the clarifications!

andreaangiolillo (approved):
LGTM. Thanks for all the changes

wtrocki:
Linting timeout is on the edge of the changes. Increasing it

wtrocki:
It looks like golangci lint uses cache for the builds. So subsequent rebuilds use previous cache. 
This is special PR that has a lot of changes so issues with timeouts should not happen with regular PRs.

wtrocki:
It looks like golangci lint uses cache for the builds. So subsequent rebuilds use previous cache. 
This is special PR that has a lot of changes so issues with timeouts should not happen with regular PRs.

PR #415, opened 2023-03-17
Demo generator cli

wtrocki:
This PR is just an experiment to show how store/unit tests and any other usage can be automatically generated with just single template. Generated files can be then put into different package or adjusted. 

This would give us out of the box store mechanism that self updates (rather than manual work when client recieves new changes)

### DIFF
--- a/mongodbatlasv2/api_organizationsatlas_stores.go
+++ b/mongodbatlasv2/api_organizationsatlas_stores.go
@@ -1,0 +1,56 @@
+ package latest
+
+import (
+	"context"
+	openapiclient "github.com/mongodb/go-client-mongodb-atlas"
+)
+
+type OrganizationsApiStore struct {
+}
+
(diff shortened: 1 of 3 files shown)


mongodbatlasv2/api_organizationsatlas_stores.go
```diff
@@ -0,0 +1,56 @@
+ package latest
+
+import (
+	"context"
+	openapiclient "github.com/mongodb/go-client-mongodb-atlas"
+)
+
+type OrganizationsApiStore struct {
```
wtrocki, line 8: That is just example of 44 different stores that were created automatically from client

tools/config/go-templates/api_store.mustache
```diff
@@ -0,0 +1,18 @@
+ package {{packageName}}
```
wtrocki, line 1: That is what is generating those stores. It is basically low effort late evening work.
If we were put more into it we can certainly make it much beter.

PR #418, opened 2023-03-21
CLOUDP-161428: Remove "Get" from getters

andmatei:
## Description
According the [Effective Go](https://go.dev/doc/effective_go#Getters) doc, getters should not contain "Get".
First commit changes the templates, second commit regenerates the client.

### DIFF
--- a/tools/config/go-templates/model_simple.mustache
+++ b/tools/config/go-templates/model_simple.mustache
@@ -90,14 +90,16 @@ func New{{classname}}WithDefaults() *{{classname}} {
 
 {{#vars}}
 {{#required}}
-// Get{{name}} returns the {{name}} field value
+{{! X-XGEN-MODIFIED }}
+// {{name}} returns the {{name}} field value
 {{#isNullable}}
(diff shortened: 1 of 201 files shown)


tools/config/go-templates/model_simple.mustache
```diff
@@ -90,14 +90,16 @@ func New{{classname}}WithDefaults() *{{classname}} {
 
 {{#vars}}
 {{#required}}
-// Get{{name}} returns the {{name}} field value
+{{! X-XGEN-MODIFIED }}
```
andreaangiolillo, line 93: What is this for?

andreaangiolillo (approved):
LGTM

tools/config/go-templates/model_simple.mustache
andmatei: Just to keep track of what we have modified in the templates, so we can easily merge the templates from upstream with our modifications.

wtrocki:
Actually. This seems like it's not going to work. It breaks linting (name will overlap). Let's close and refine our strategy in what methods we want in model.

PR #424, opened 2023-03-23
Remove unused model template from go templates

wtrocki:
Removing file that is not used.


matt-condon (approved):
🙏

wtrocki:
Merging as golangcilint problem is related with github issue. Change doesn't touch any go code.

PR #427, opened 2023-03-23
CLOUDP-168044: disable readonly models marshaling

wtrocki:
## Description

CLOUDP-168044

Disable read-only models marshaling functionality. 
Readonly models are by default removed from JSON string to prevent them from being sent to the server.

Only first commit matters. Rest of the changes are regenerated models (almost every response type.

## Implications
 
This change has some implications that we need to test
If the model is used in the request and response feature was removing read-only fields in a request.
For example ID of the object was removed automatically. When this is missing we need to test if extra objects not causing any side effects.

### DIFF
--- a/mongodbatlasv2/model_access_list_item.go
+++ b/mongodbatlasv2/model_access_list_item.go
@@ -99,18 +99,15 @@ func (o *AccessListItem) SetIpAddress(v string) {
 	o.IpAddress = v
 }
 
-func (o AccessListItem) MarshalJSON() ([]byte, error) {
+func (o AccessListItem) MarshalJSONWithoutReadOnly() ([]byte, error) {
 	toSerialize,err := o.ToMap()
 	if err != nil {
 		return []byte{}, err
 	}
 	return json.Marshal(toSerialize)
 }
-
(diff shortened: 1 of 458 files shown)


matt-condon (approved):
Nice. LGTM!

gssbzn:
>If the model is used in the request and response

We do this a lot

> When this is missing we need to test if extra objects not causing any side effects.

A lot of APIs throw errors, check https://github.com/mongodb/mongodb-atlas-cli/blob/1bd6915dff43ca50ca930fea14c1c3227fdb3b42/internal/cli/atlas/clusters/clusters.go#L126-L150

wtrocki:
I'm investigating now solution when:

- Marshal method gets removed as what it would do now is 
- A retrofit method that removes optional from object same as pointed by @gssbzn

wtrocki:
I think that with the extend of changes we making we generally improving our user facing API but in result we driving even further from OpenAPI generator templates so our updates will become much harder. 
CC @andmatei

wtrocki:
Basically, renaming method enables us to achieve both:

- We can remove arguments with the new method if it is needed (however underlying client doesn't seem to use that anyway).

Verified with examples. All seems to be working fine.

PR #302, opened 2022-07-01
CLOUDP-127482: Add filter by name on list all groups by org

fmenezes:
## Description

Add filter by name on list all groups by org

Link to any related issue(s): CLOUDP-127482

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Required Checklist:

- [ ] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added any necessary documentation (if appropriate)
- [x] I have run `make fmt` and formatted my code

PR #436, opened 2023-03-27
POC: add ListOptions to endpoints that return paginated views

andmatei:


tools/config/go-templates/api.mustache
```diff
@@ -59,6 +59,16 @@ type {{#structPrefix}}{{&classname}}{{/structPrefix}}{{^structPrefix}}Api{{/stru
 {{/allParams}}
 }
 
+{{#vendorExtensions.x-xgen-paginated}}
+func (r {{#structPrefix}}{{&classname}}{{/structPrefix}}{{^structPrefix}}Api{{/structPrefix}}{{operationId}}Request) ListOptions(listOptions *ListOptions) {{#structPrefix}}{{&classname}}{{/structPrefix}}{{^structPrefix}}Api{{/structPrefix}}{{operationId}}Request {
+	r.pageNum = &listOptions.PageNum
```
andmatei, line 64: These are hard-coded for this POC, but can auto-generated in the future based on other extensions

wtrocki (approved):
Invalid

openapi/atlas-api-transformed.yaml
```diff
@@ -11370,6 +11370,7 @@ paths:
       summary: Return All Serverless Instances from One Project
       tags:
         - Serverless Instances
+      x-xgen-paginated: "true"
```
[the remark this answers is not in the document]
wtrocki: The same can be done by the interface. Defining APIList interface within API.

tools/config/go-templates/atlas_client.mustache
```diff
@@ -91,3 +91,16 @@ func UseUserAgent(userAgent string) ClientModifier {
 		return nil
 	}
 }
+
+// ListOptions specifies the optional parameters to List methods that
+// support pagination.
+type ListOptions struct {
```
wtrocki, line 97: atlas_client typically should not contain objects that are subject to api changes. 
Typical pattern in openapi generator is to use:

- Models (for all in and out parameters)
- API (for way to interact with those)

That being said this looks like model although it does not come from the OpenAPI, which significantly affects our ability to autogenerate future versions that might introduce improvements in the List operations (API is where those shouldt typically happen).

wtrocki:
I like idea but it needs more refinement. If your target is to aid other clients with some helpful structures the easiest way would be to add them as static files to the package. However those cannot have json parser (should not be used in requests to not send wrong message and break clients when change in lists will happen)

openapi/atlas-api-transformed.yaml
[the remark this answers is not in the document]
andmatei: I do not really follow this. Could you please elaborate more? What are you referring to by APIList interface?

tools/config/go-templates/atlas_client.mustache
andmatei: This is just a POC to show what could be possible. Ideally this would be auto-generated based on the properties of our Paginated views (as per my comment above). Will invest some learning time to see how to make this through generation.

mongodbatlasv2/atlas_client.go
```diff
@@ -91,3 +91,16 @@ func UseUserAgent(userAgent string) ClientModifier {
 		return nil
 	}
 }
+
+// ListOptions specifies the optional parameters to List methods that
+// support pagination.
+type ListOptions struct {
```
gssbzn, line 97: who is the user of this struct?
what do they gain from this?

tools/config/go-templates/atlas_client.mustache
wtrocki: My point is that it doesn't need to be generated. It can be helper that is separate to the generate SDK (and should not confuse users that this is model or some form of api contract as it will seriousy restrict us in the futrue.

mongodbatlasv2/atlas_client.go
andmatei: I would say we could would be one of the users for this, for the AtlasCLI migration project as it would result in less work in refactoring the list store methods, i.e., 
```
func (s *Store) Projects(opts *atlas.ListOptions) 
```
would become
```
func (s *Store) Projects(opts *atlasv2.ListOptions) 
```
instead of deciding on a new pattern for this methods across AtlasCLI.

wtrocki: I'm fine for us to create a list object but it should not be used as a structure that is marshaled/used as model. 
Using templates for helper structure is something that I will allow, but I would rather us to write go code due to degraded maintenance for template form.

gssbzn: >for the AtlasCLI migration project as it would result in less work in refactoring the list store methods

Then why don't we add the code on the CLI?

wtrocki: ⬆️

andmatei: > Then why don't we add the code on the CLI?
Seemed more like a SDK functionality than a CLI one, since it would also be easier for customers to migrate to the new SDK with list options

> How would you handle cases where list includes more than pagination, things like filters, for example metrics?
If we stick with this, the other parameters could be specified through the current mechanism we have (chaining the parameters on the request). Otherwise, we could have different levels of parameter grouping, that would result in ListOptions for each list method (i.e., `OrganizationListOptions` , `MetricsListOptions` etc.)

gssbzn: Food for though, the new SDK uses a fluent interface, `GetThinh().Page().Limit().OtherThing().Execute()` fluent APIs are great and it saves us from a trap we felt (I say "we" even the initial implementation predates anyone in the team) by using a more imperative approach `GetThing(arg1, arg2, argN)`, why a trap, check https://github.com/mongodb/go-client-mongodb-atlas/pull/302, when we get the query params wrong it tends to be a breaking change, now this is why I think the new way is better than the current way and `ListOptions` should be a thing of the past

Now about this PR, it comes with a support cost, once we add it we can't delete, how long do we expect users to use it as a stepping stone to migrate to the new SDK? what do we do if a support case comes in and asks for the equivalent for all related list options since we have a precedence of adding this method here so a bit of a slippery slope

TL;DR I fear adding it here may come with hidden costs, adding to the CLI may be ok, other users may follow a similar approach of internally managing transformations and that's ok, I belive but may need to double check we set a clear goal of not making the new SDK a straight up replacement so we are safe on that side as well

wtrocki: Awesome response! Thank you @gssbzn

PR #438, opened 2023-03-28
CLOUDP-168427: remove parameters from the Error object

wtrocki:
CLOUDP-168427

## Description

Temp workaround that will enable us to process errors in the GO SDK. 
Actuall fix made in the openapi.

### DIFF
--- a/openapi/atlas-api-transformed.yaml
+++ b/openapi/atlas-api-transformed.yaml
@@ -21835,9 +21835,7 @@ components:
           example: TOO_MANY_GROUP_NOTIFICATIONS
         parameters:
           type: array
-          items:
-            type: object
-            description: Parameter uses to give more information about the error.
+          items: {}
         reason:
           type: string
           description: Application error message returned with this error.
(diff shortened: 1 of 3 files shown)


openapi/atlas-api-transformed.yaml
```diff
@@ -21833,11 +21833,6 @@ components:
           type: string
           description: Application error code returned with this error.
           example: TOO_MANY_GROUP_NOTIFICATIONS
-        parameters:
-          type: array
-          items:
-            type: object
```
gssbzn, line 21839: reading the ticket I guess the problem comes because we define this an as array of object, but what we mean is an array of "anything", being an array of object openapi-generator is assuming `[]map[string]interface{}` but it should be `[]interface{}` what's the best way to model this on openapi?

wtrocki: My line of thinking - what is the value of that in error. 
We are returning parameters that user sent?

wtrocki: Ok. Ignore me

wtrocki:
Fixed this following open API spec. I think that the common pattern for an array returning everything is to return `{}`.
Returning object array strongly implies map.

gssbzn (approved):
LGTM

PR #443, opened 2023-03-30
CLOUDP-167160: generic feature. Enable default values for query param…

wtrocki:
## Description

We noticed that v2 sdk query parameters do not receive default values. 
This PR fixes that on our SDK. We need to backport that change to openapi generator later phase.

For review see: tools/config/go-templates/api.mustache
I did tested it with example - seems like defaults are taken into account.

### DIFF
--- a/openapi/atlas-api-transformed.yaml
+++ b/openapi/atlas-api-transformed.yaml
@@ -21769,6 +21769,7 @@ components:
           example: TOO_MANY_GROUP_NOTIFICATIONS
         parameters:
           type: array
+          description: Parameter uses to give more information about the error.
           items: {}
         reason:
           type: string
--- a/tools/config/go-templates/api.mustache
+++ b/tools/config/go-templates/api.mustache
@@ -219,8 +219,7 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 	{{#defaultValue}}} else {
 		var defaultValue {{{dataType}}} = {{{.}}}
 		r.{{paramName}} = &defaultValue
-	{{/defaultValue}}
-	}
+	{{/defaultValue}}}
 	{{/required}}
 	{{/queryParams}}
 	// to determine the Content-Type header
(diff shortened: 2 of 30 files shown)


openapi/atlas-api-transformed.yaml
```diff
@@ -21769,6 +21769,7 @@ components:
           example: TOO_MANY_GROUP_NOTIFICATIONS
         parameters:
           type: array
+          description: Parameter uses to give more information about the error.
```
gssbzn, line 21772: 👀  I know this is a thing on the spec but I guess this should be 
```suggestion
          description: Parameter used to give more information about the error.
```

tools/config/go-templates/api.mustache
```diff
@@ -216,7 +216,10 @@ func (a *{{{classname}}}Service) {{nickname}}Execute(r {{#structPrefix}}{{&class
 	{{^isCollectionFormatMulti}}
 		parameterAddToHeaderOrQuery(localVarQueryParams, "{{baseName}}", r.{{paramName}}, "{{collectionFormat}}")
 	{{/isCollectionFormatMulti}}
-	}
+	{{#defaultValue}}} else {
+		var defaultValue {{{dataType}}} = {{{.}}}
+		r.{{paramName}} = &defaultValue
+	{{/defaultValue}}}
```
gssbzn, line 222: just for me to understand, why do we need to set the default? I was assuming the problem was because of not using pointers, and in which case it was defaulting to false. I would expect that while using a pointer and having the omit empty instruction the parameter won't get serialized then default to what the API defaults, is this not how it works for the generated SDK?

gssbzn (approved):
LGTM

tools/config/go-templates/api.mustache
wtrocki: I hope I understand question:

Query parameters aren't serialized - they are parsed into request url.
Required query params are always on method to create request. 
Optional ones are in struct and pointers which might be confusing.

Adding default support gives us ability to improve contract for optional parameters (omit empty does not apply).
The rest is like you have mentioned - if someone did not provided value we do check if default is present and assign it.

openapi/atlas-api-transformed.yaml
wtrocki: Created jira to fix upstream

tools/config/go-templates/api.mustache
gssbzn: >Query parameters aren't serialized 

Ok this may be the gap I may have, we were using a google lib that let's you define query params the same as a you do for json, and it applies the same rules as for json and nnotations, https://github.com/google/go-querystring seems like I need to give how openapi-gen does this a look to get familiar with it

PR #451, opened 2023-04-13
chore: reintroduce transformer workarounds

wtrocki:
## Description

Changes are functional in the dev branch as we tested, but they still not available in production.
We need to keep workarounds for some time. I created jira to remove them in GA epic.

### DIFF
--- a/tools/transformer/src/atlasTransformations.js
+++ b/tools/transformer/src/atlasTransformations.js
@@ -39,6 +39,22 @@ module.exports = function runTransformations(openapi) {
     "View",
     ignoredModelNames
   );
+  
+  // Temp workaround for
+  // https://jira.mongodb.org/browse/CLOUDP-166120
+  openapi.components.responses.noBody = {
+    content: {
+      "application/vnd.atlas.2023-01-01+json": {
+        example: "",
+      },
+    },
+    description: "This endpoint does not return a response body",
+  };
+
+  // Temp workaround for CLOUDP-168427
+  if (openapi.components.schemas.Error) {
+    openapi.components.schemas.Error.properties.parameters.items = {};
+  }
 
   return openapi;
 };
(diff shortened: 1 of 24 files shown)


tools/transformer/src/atlasTransformations.js
```diff
@@ -40,5 +40,21 @@ module.exports = function runTransformations(openapi) {
...
+        example: "",
+      },
+    },
+    description: "This endpoint does not return a response body",
+  };
+
+  // Temp workaround for CLOUDP-168427
+  if (openapi.components.schemas.Error) {
+    openapi.components.schemas.Error.properties.parameters.items = {};
+  }
```
gssbzn, line 57: why haven't we fixed this upstream?

wtrocki: We did but we are waiting for fix to arrive in 3 weeks.

tools/transformer/src/atlasTransformations.js
```diff
@@ -40,5 +40,21 @@ module.exports = function runTransformations(openapi) {
...
+  // Temp workaround for
+  // https://jira.mongodb.org/browse/CLOUDP-166120
+  openapi.components.responses.noBody = {
+    content: {
+      "application/vnd.atlas.2023-01-01+json": {
+        example: "",
+      },
+    },
+    description: "This endpoint does not return a response body",
+  };
```
gssbzn, line 52: this was resolved upstream but missed deployment, can you please backport @matt-condon

wtrocki: No need for backport as we are using dev version of the sdk for that work with other fixes. 
But since both dev and prod use the same codebase it is better to keep workarounds for little bit longer. 
I have jira to remove them so we good.

gssbzn: do you have the ticket?

tools/transformer/src/atlasTransformations.js
gssbzn: do you have a ticket on the fix upstream?

wtrocki:
Side note. I have reintroduced previously removed fixes.

matt-condon (approved):

PR #455, opened 2023-04-14
Remove unnedded workflow

wtrocki:
Remove workflow that is no longer used and documented. 
That workflow was initially setup for testing. RIght now we rely on dev and production workflows that are officially documented. This one were triggered only when sdkv2 recieves updates (which will create a lot of redundant runs after OpenAPI PR is merged)


gssbzn (approved):
LGTM

PR #456, opened 2023-04-14
Add error handling for SDK

wtrocki:
## Description

Add error handling for the sdk with documentation

### DIFF
--- a/sdk/errors.go
+++ b/sdk/errors.go
@@ -6,8 +6,8 @@ import (
 	mongodbatlasv2 "go.mongodb.org/atlas/mongodbatlasv2"
 )
 
-// GetAPIError gets a strongly typed error from an error.
-func GetAPIError(err error) *mongodbatlasv2.Error {
+// APIError gets a strongly typed error from an error.
+func APIError(err error) *mongodbatlasv2.Error {
 	var openapiError mongodbatlasv2.GenericOpenAPIError
 
 	if ok := errors.As(err, &openapiError); !ok {
--- a/sdk/errors.go
+++ b/sdk/errors.go
@@ -11,7 +11,7 @@ func APIError(err error) *mongodbatlasv2.Error {
 	var openapiError mongodbatlasv2.GenericOpenAPIError
 
 	if ok := errors.As(err, &openapiError); !ok {
-		return nil	
+		return nil
 	}
 	errModel := openapiError.Model()
 	transformedError, ok := errModel.(mongodbatlasv2.Error)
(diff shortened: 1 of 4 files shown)


sdk/errors.go
```diff
@@ -0,0 +1,35 @@
+package sdk
+
+import (
+	"errors"
+
+	mongodbatlasv2 "go.mongodb.org/atlas/mongodbatlasv2"
+)
+
+// GetAPIError gets a strongly typed error from an error.
+func GetAPIError(err error) *mongodbatlasv2.Error {
```
gssbzn, line 10:
```suggestion
// APIError gets a strongly typed error from an error.
func APIError(err error) *mongodbatlasv2.Error {
```
[no getters](https://go.dev/doc/effective_go#Getters)

sdk/errors.go
```diff
@@ -0,0 +1,35 @@
...
+		errModel := openapiError.Model()
+
+		transformedError, ok := errModel.(mongodbatlasv2.Error)
+		if !ok {
+			return nil
+		}
+		return &transformedError
+	}
+
+	return nil
```
gssbzn, line 23: [style] I prefer early returns as they avoid deep nesting
```suggestion
	if ok := errors.As(err, &openapiError); !ok {
		return nil	
	}
	errModel := openapiError.Model()
	transformedError, ok := errModel.(mongodbatlasv2.Error)
	if !ok {
		return nil
	}
	return &transformedError
```

sdk/errors.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// IsAPIError returns true if the error contains the errCode.
+// Error code is an code that is returned by the API.
+func IsAPIError(err error, code string) bool {
```
gssbzn, line 28:
```suggestion
func IsErrorCode(err error, code string) bool {
```

gssbzn (commented):
some small  style comments and a question, is the goal to use sdk for the util functions?

wtrocki:
> some small style comments and a question, is the goal to use sdk for the util functions?

Yes. 
I struggled with naming so if you have any proposition let me know. 
We need place to put some non generated code like errors, pagination helpers etc.

CC @andmatei :godmode:

gssbzn:
> I struggled with naming so if you have any proposition let me know.

Up to me a package errors for these methods, a package pagination for pagination and so on to follow a good practice of meaningful names instead of generic `utils` packages

gssbzn (approved):

wtrocki:
My worry is that without pkg folder we going to end up with a lot of folders containing 1-2 go files vs single client folder containing +500 that is hard to find. If we introduce pkg then mongodbatlasv2 would not be appropriate. 
Introducing some subfolder of the generated code will complicate sdk cleanup. 

If possible will look to improve this for GA (were we going to get separate repository so layout can change)

wtrocki:
Will merge but if you have any opinion I'm happy to fix this in follow up PR.

PR #459, opened 2023-04-19
fix: enable casting API errors into error object

wtrocki:
## Description

Extending SDK with another useful helper method that will map API error objects into generic error that CLI, Terraform might use etc.


sdk/errors.go
```diff
@@ -21,6 +22,15 @@ func APIError(err error) *mongodbatlasv2.Error {
 	return &transformedError
 }
 
+// APIErrorToErr returns APIError as instance of Go error or nil if passed error is not an API error.
+func APIErrorToErr(err error) error {
+	apiError := APIError(err)
+	if apiError != nil {
+		return fmt.Errorf("%v", apiError)
```
wtrocki, line 29: we need this for atlas cli that returns generic errors.
Alternatively we can use Wrap/Unwrap to carry original error but it will also drive a lot of boilerplate.

ApiError can be also stringified to JSON string

gssbzn (changes requested):
can APIError implement `Error` so it fullfils the `error` interface?, ideally, in  backward compatible way? https://github.com/mongodb/go-client-mongodb-atlas/blob/437d9b43222e10c0fd565d979eafd85f80c199e2/mongodbatlas/mongodbatlas.go#L539-L542

wtrocki:
> can APIError implement Error so it fullfils the error interface?, ideally, in backward compatible way?

Very good question. I thought of having error object defined separately (not as model)
As for now API error is tied to the OpenAPI thus it is just an ordinary model. 
However our contracts for errors are very weak. I'm thinking what would be the best way forward in this case.

wtrocki:
I have the solution! IMHO we can move all error handling methods to client - thus we would not need to have different imports and can extend existing structures (subject to volatility if we were to change error object name in future)

wtrocki:
So generally based on our architecture Error object cannot be error. It needs to be API object only and it is always wrapped by GenericError that has more metadata and it is error object itself. 

I have made GenericError object strongly typed now and rely 100% on Error object.

Both formats work out of the box right now:

err.Error() - used in CLI to print message prints backwards compatible message (with some extended content)
v2.IsError and v2.IsErrorCode() give us error structures and way to handle specific errors. 

With both in place we should have complete error suite. 
Good news is that this change doesn't require any adjustment in Atlas CLI and it will just work.

mongodbatlasv2/atlas_client.go
```diff
@@ -91,3 +92,22 @@ func UseUserAgent(userAgent string) ClientModifier {
...
+
+// IsError checks if API returned known error type.
+func IsError(err error) (*Error, bool) {
+	var openapiError *GenericOpenAPIError
+	if ok := errors.As(err, &openapiError); !ok {
+		return nil, false
+	}
+	errModel := openapiError.Model()
+	return &errModel, true
+}
```
gssbzn, line 104: is this method necessary now that `GenericOpenAPIError` always returns the correct struct and there's no need for casting?

mongodbatlasv2/atlas_client.go
```diff
@@ -91,3 +92,22 @@ func UseUserAgent(userAgent string) ClientModifier {
...
+
+// IsErrorCode returns true if the error contains the specific code.
+func IsErrorCode(err error, code string) bool {
+	mappedErr, _ := IsError(err)
+	if mappedErr == nil {
+		return false
+	}
+	return mappedErr.GetErrorCode() == code
+}
```
gssbzn, line 113: should this be 
```suggestion
func (e GenericOpenAPIError) IsErrorCode(code string) bool {
	return e.GetError().GetErrorCode() == code
}
```

mongodbatlasv2/atlas_client.go
wtrocki: SDK has 2 levels of errors. Internal errors like io/output errors/networking issues etc. Actual API errors represented by GenericOpenAPIError. We need to handle the boilerplate of casting to GenericOpenAPIError to avoid documentation (and general confusion related to how errors are handled.

mongodbatlasv2/atlas_client.go
wtrocki: I think that GenericOpenAPIError is an implementation detail and should never be part of the interface (used by the clients). Users interact with 2 methods that are fairly generic and return only what they want.

tools/config/go-templates/atlas_client.mustache
```diff
@@ -91,3 +92,22 @@ func UseUserAgent(userAgent string) ClientModifier {
 		return nil
 	}
 }
+
+// IsError checks if API returned known error type.
+func IsError(err error) (*Error, bool) {
```
gssbzn, line 97: let's follow similar semantics to the std lib 
```suggestion
// AsError  if err matches Error it returns its value and true. Otherwise, it returns nil and false.
func AsError(err error) (*Error, bool) {
```

wtrocki: Good point. Is is for booleans :(

PR #461, opened 2023-04-20
CLOUDP-167941 - Remove Regression Tests and Pointy integration 

wtrocki:
## Description

Regression tests will likely be broken when new versions of the SDK will be applied. 
They served us well for the initial period of the project. Since now we have versioned API and breaking changes support regression tests became stale. As result we also remove Pointy 🙈


gssbzn (approved):
LGTM

PR #463, opened 2023-04-21
fix: externalise structure for query and path params

wtrocki:
## Description

Create separate structure for query params

## Long

As part of the migration, I see recurrent theme where GET url arguments are returned as part of the Fluent API. E.g
 ```
 s.clientv2.PerformanceAdvisorApi.
		ListSuggestedIndexes(s.ctx, projectID, processName).
		NExamples(1) // This is query param
		Execute()		
```	
Due to the architecture in the CLI (stores) we see need to pass those arguments to functions
This usually means:
- Passing from 3 to 10 arguments to methods
- Create some intermediate objects in the CLI

Manual SDKs have been supporting this by GET arguments being represented as struct:
https://github.com/mongodb/go-client-mongodb-atlas/blob/437d9b43222e10c0fd565d979eafd85f80c199e2/mongodbatlas/performance_advisor.go#L105-L130

This PR brings parity (with level of backwards compatibility between both SDKs.

### DIFF
--- a/mongodbatlasv2/api_access_tracking.go
+++ b/mongodbatlasv2/api_access_tracking.go
@@ -71,7 +71,7 @@ type AccessTrackingApiListAccessLogsByClusterNameRequest struct {
 	start *time.Time
 }
 
-type AccessTrackingApiListAccessLogsByClusterNameQueryParams struct {
+type AccessTrackingApiListAccessLogsByClusterNameParams struct {
 		GroupId string
 		ClusterName string
 		AuthResult *bool
@@ -262,7 +262,7 @@ type AccessTrackingApiListAccessLogsByHostnameRequest struct {
 	start *time.Time
 }
 
-type AccessTrackingApiListAccessLogsByHostnameQueryParams struct {
+type AccessTrackingApiListAccessLogsByHostnameParams struct {
 		GroupId string
 		Hostname string
 		AuthResult *bool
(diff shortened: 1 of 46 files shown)

### CODE REVIEW

File: mongodbatlasv2/api_x509_authentication.go [hunk 1]
```diff
@@ -332,6 +340,13 @@ type X509AuthenticationApiListDatabaseUserCertificatesRequest struct {
 	itemsPerPage *int32
 	pageNum *int32
 }
+type X509AuthenticationApiListDatabaseUserCertificatesQueryParams struct {
+		groupId string
+		username string
+		includeCount *bool
+		itemsPerPage *int32
+		pageNum *int32
```
gssbzn, line 348: fields are private so they can't be used

wtrocki: Moving to draft. Sorry for early ping

mongodbatlasv2/api_access_tracking.go
```diff
@@ -71,6 +71,16 @@ type AccessTrackingApiListAccessLogsByClusterNameRequest struct {
 	start *time.Time
 }
 
+type AccessTrackingApiListAccessLogsByClusterNameQueryParams struct {
+		GroupId string
+		ClusterName string
```
gssbzn, line 76: group id and cluster name are path param, is this the expected usage?

wtrocki: Yes. We need all arguments (file path params and query params in one place. 
I did failed naming the structure correctly thought. 
Will use `Opts` to make it more backwards compatible with v1

mongodbatlasv2/api_access_tracking.go
```diff
@@ -71,7 +71,7 @@ type AccessTrackingApiListAccessLogsByClusterNameRequest struct {
 	start *time.Time
 }
 
-type AccessTrackingApiListAccessLogsByClusterNameQueryParams struct {
+type AccessTrackingApiListAccessLogsByClusterNameParams struct {
```
wtrocki, line 74: Since we guarantee the uniqueness of the operationIds I belive that we can remove first suffix as well

gssbzn (approved):
LGTM

PR #464, opened 2023-04-21
Enable fluent and struct API for go SDK

wtrocki:
One of the biggest problems integrations might face is that fluent API forces end users to use specific interface (chain of operations/builder pattern-like) solutions. In many cases, this will mean that migration between SDKv1 and SDKv2 will be quite painful.

This PR introduces a new method to the API object that can accept structs. 

As result, users will get:

- Clean, effortless migration path from v1 to v2 (removing a lot of manual steps and a complete rewrite of the data layer or even the application layer)
- Access to individual resources without the need to duplicate all requirement
- Simple means to understand what arguments are required for each API method (right now users need to discover various objects. 

While this change might look big due to the number of files changed what we changing here is only a single template that adds a new method.
I will be looking to adopt this upstream as well. Change is not breaking - it doesn't affect any goals and existing documentation etc.

## Notes about recent updates

1. I have made this feature experimental for now. Fluent API is still the default feature
2. I will focus on small breaking changes for fluent API in separate PR to simplify the review. 
This PR contains all non-breaking changes for atlas CLI (tested end to end)

### DIFF
--- a/examples/example_cluster_aws.go
+++ b/examples/example_cluster_aws.go
@@ -37,7 +37,7 @@ func main() {
 
 	// -- 1. Get first project
 	projects, response, err := sdk.ProjectsApi.ListProjects(ctx).
-		IncludeCount(false).ItemsPerPage(1).Execute()
+		IncludeCount(false).Execute()
 	handleErr(err, response)
 
 	if projects.GetTotalCount() == 0 {
--- a/examples/example_cluster_aws.go
+++ b/examples/example_cluster_aws.go
@@ -35,9 +35,14 @@ func main() {
 		mongodbatlas.UseDebug(false))
 	handleErr(err, nil)
 
-	// -- 1. Get first project
-	projects, response, err := sdk.ProjectsApi.ListProjects(ctx).
-		IncludeCount(false).Execute()
+	// -- 1. Get first project using fluent API
+	_, response, err := sdk.ProjectsApi.ListProjects(ctx).
+		IncludeCount(false).ItemsPerPage(1).Execute()
+	handleErr(err, response)
+
+	// -- 1.1 Get first project using struct based API
+	listParams := &mongodbatlas.ListProjectsApiParams{ItemsPerPage: mongodbatlas.PtrInt32(1)}
+	projects, response, err := sdk.ProjectsApi.ListProjects(ctx).ExecuteWithParams(listParams)
 	handleErr(err, response)
 
 	if projects.GetTotalCount() == 0 {
--- a/examples/example_cluster_aws.go
+++ b/examples/example_cluster_aws.go
@@ -42,7 +42,7 @@ func main() {
 
 	// -- 1.1 Get first project using struct based API
 	listParams := &mongodbatlas.ListProjectsApiParams{ItemsPerPage: mongodbatlas.PtrInt32(1)}
-	projects, response, err := sdk.ProjectsApi.ListProjects(ctx).ExecuteWithParams(listParams)
+	projects, response, err := sdk.ProjectsApi.ListProjectsWithParams(ctx, listParams).Execute()
 	handleErr(err, response)
 
 	if projects.GetTotalCount() == 0 {
--- a/examples/example_cluster_aws.go
+++ b/examples/example_cluster_aws.go
@@ -35,14 +35,9 @@ func main() {
 		mongodbatlas.UseDebug(false))
 	handleErr(err, nil)
 
-	// -- 1. Get first project using fluent API
-	_, response, err := sdk.ProjectsApi.ListProjects(ctx).
+	// -- 1. Get first project
+	projects, response, err := sdk.ProjectsApi.ListProjects(ctx).
 		IncludeCount(false).ItemsPerPage(1).Execute()
-	handleErr(err, response)
-
-	// -- 1.1 Get first project using struct based API
-	listParams := &mongodbatlas.ListProjectsApiParams{ItemsPerPage: mongodbatlas.PtrInt32(1)}
-	projects, response, err := sdk.ProjectsApi.ListProjectsWithParams(ctx, listParams).Execute()
 	handleErr(err, response)
 
 	if projects.GetTotalCount() == 0 {
--- a/mongodbatlasv2/api_access_tracking.go
+++ b/mongodbatlasv2/api_access_tracking.go
@@ -36,7 +36,7 @@ type AccessTrackingApi interface {
 
 	@param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
 	@param ListAccessLogsByClusterNameApiParams - Parameters for the request
-	@return ListAccessLogsByClusterNameApiRequest}}
+	@return ListAccessLogsByClusterNameApiRequest
 	*/
 	ListAccessLogsByClusterNameWithParams(ctx context.Context, args *ListAccessLogsByClusterNameApiParams) ListAccessLogsByClusterNameApiRequest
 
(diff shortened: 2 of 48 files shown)


wtrocki:
As an outstanding item, we might need to clarify what to do with not optional arguments (Those are now supplied to method that creates requests and cannot be avoided. I really wanted to avoid breaking change for fluent API

wtrocki:
Biggest challenge/inconvenience of that approach is that we need to supply an struct to the API containing nonoptional fields that we already supplied, making it an bad api. We can eventually generate separate struct for that purpose but it would seem to be counterintuitive.

mongodbatlasv2/api_alert_configurations.go
```diff
@@ -708,6 +730,14 @@ func (r ListAlertConfigurationsApiRequest) Execute() (*PaginatedAlertConfig, *ht
 	return r.ApiService.ListAlertConfigurationsExecute(r)
 }
 
+func (r ListAlertConfigurationsApiRequest) ExecuteWithParams(params *ListAlertConfigurationsApiParams) (*PaginatedAlertConfig, *http.Response, error) {
+	r.groupId = params.GroupId 
+	r.includeCount = params.IncludeCount 
+	r.itemsPerPage = params.ItemsPerPage 
+	r.pageNum = params.PageNum
```
gssbzn, line 737: how would this behave for `includeCount`
```go
params := ListAlertConfigurationsApiParams{
  pageNum: 1,
}
```

I expect the response to include count because include count default on the API is true

gssbzn: this to align on the goal of 
>Clean, effortless migration path from v1 to v2

Since v1 relies on the traditional serialization logic for optional fields and omitting empty

mongodbatlasv2/api_access_tracking.go
```diff
@@ -115,6 +115,17 @@ func (r ListAccessLogsByClusterNameApiRequest) Execute() (*MongoDBAccessLogsList
 	return r.ApiService.ListAccessLogsByClusterNameExecute(r)
 }
 
+func (r ListAccessLogsByClusterNameApiRequest) ExecuteWithParams(params *ListAccessLogsByClusterNameApiParams) (*MongoDBAccessLogsList, *http.Response, error) {
+	r.groupId = params.GroupId 
+	r.clusterName = params.ClusterName 
+	r.authResult = params.AuthResult
```
gssbzn, line 121: similar to https://github.com/mongodb/go-client-mongodb-atlas/pull/464/files#r1174923339 this is a perfect case to test this PR
`authResult` when `true` means successful access, `false` means unsuccessful access and not set means both, can this method needs to cover the three cases

wtrocki: In nonatlas specific cases `default` means that *if field is not present* backend will assume that value
This means that default should never be passed from the client (and even if they are they are redundant).

I have made a recent fix to pass the default values from the client for query params (which would never have landed upstream due to the above problem). Do we want to incorporate that here as well

Since we setting values checking for non nil values makes no sense (as it will end up with the same result anyway)

wtrocki: Let's clarify it because we might have misunderstood what default behavior we want from client. 
For me, default is clearly informational for the end user and should not be present in the client.
The reality is that our defaults might not be applied by some of the atlas endpoints right? so that is why we need to pas them?

wtrocki: Done by enabling *bool values:

- true,false
- Nil (apply default when sending request or ignore when  omitEmpty is present

examples/example_cluster_aws.go
```diff
@@ -37,7 +37,7 @@ func main() {
 
 	// -- 1. Get first project
 	projects, response, err := sdk.ProjectsApi.ListProjects(ctx).
-		IncludeCount(false).Execute()
+		IncludeCount(false).ItemsPerPage(1).Execute()
```
gssbzn, line 40: verifying, was this change necessary? 1 should already be the default

tools/config/go-templates/api.mustache
```diff
@@ -33,13 +33,23 @@ type {{classname}} interface {
...
 
-	// {{nickname}}Execute executes the request{{#returnType}}
-	//  @return {{{.}}}{{/returnType}}
+	@param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+	@param {{operationId}}ApiParams - Parameters for the request
+	@return {{operationId}}ApiRequest}}
 	{{#isDeprecated}}
-	// Deprecated
+
+	Deprecated: Method have been deprecated. Please check the latest resource version for {{classname}}
```
gssbzn, line 46:
```suggestion
	Deprecated: this method has been deprecated. Please check the latest resource version for {{classname}}
```

mongodbatlasv2/api_access_tracking.go
```diff
@@ -30,10 +30,18 @@ type AccessTrackingApi interface {
 	@return ListAccessLogsByClusterNameApiRequest
 	*/
 	ListAccessLogsByClusterName(ctx context.Context, groupId string, clusterName string) ListAccessLogsByClusterNameApiRequest
+	/*
+	ListAccessLogsByClusterName Return Database Access History for One Cluster using Its Cluster Name
+
+
+	@param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+	@param ListAccessLogsByClusterNameApiParams - Parameters for the request
+	@return ListAccessLogsByClusterNameApiRequest}}
```
gssbzn, line 39: there seem to be some dangling braces (}) on the template

tools/config/go-templates/api.mustache
```diff
@@ -33,13 +33,23 @@ type {{classname}} interface {
...
+	{{!X-GEN-CUSTOM}}
+	/*
+	{{operationId}} {{{summary}}}{{^summary}}Method for {{operationId}}{{/summary}}
+
 
-	// {{nickname}}Execute executes the request{{#returnType}}
-	//  @return {{{.}}}{{/returnType}}
+	@param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+	@param {{operationId}}ApiParams - Parameters for the request
+	@return {{operationId}}ApiRequest}}
```
gssbzn, line 43:
```suggestion
	@return {{operationId}}ApiRequest
```

mongodbatlasv2/api_access_tracking.go
gssbzn: >Nil (apply default when sending request or ignore when omitEmpty is present

Do you have an example of omitempty? I can't see it on the template and can't tell if it's really in use

>Done by enabling *bool

What about string("") and numbers(0)

examples/example_cluster_aws.go
wtrocki: Default is 100. Page number is 1. Removing it as it doesn't seem to make much difference in example

gssbzn:
waiting on https://github.com/mongodb/go-client-mongodb-atlas/pull/464#discussion_r1188312909

mongodbatlasv2/api_access_tracking.go
wtrocki: > Do you have an example of omitempty? I can't see it on the template and can't tell if it's really in use

My bad.
Omit empty would be only for request body objects. Params objects are intermediate objects to hold arguments, they aren't used in serialization.

> What about string("") and numbers(0)

String and numbers are pointers (as internal structs also use pointers), thus solving the problem of uncertainty if the value is provided. There would be also outstanding work on the atlas cli to have flags that do provide empty values (instead of 0 or math.MaxInt32 it can be nil)

gssbzn (approved):
LGTM thanks for all the work here

PR #468, opened 2023-05-05
test: update actions to use the new go action values

gssbzn:
## Description

Make go tests use stable and previous stable as versions so we can make these variants mandatory on PRs


wtrocki:
Will bring that back to sdkv2

PR #470, opened 2023-05-08
CLOUDP-175598: openapi pipeline to run client test

wtrocki:
## Description

Example application is subject to be broken when a new Resource version of the API will be introduced that changes the resources it uses. That is fine for us as then update PRs will fail and will require manual correction (which we want).

However, example applications were also used as tests for CI/CD processes which is not desired breaking changes will also break the job.  This PR improves the openapi pipeline to rely on client test (it will compile entire client codebase by running the test).

## Verification done
```
make openapi_pipeline
```

Running tests with broken codebase:
```
# go.mongodb.org/atlas/mongodbatlasv2
mongodbatlasv2/api_clusters.go:87:36: undefined: CloudProviderRegionsApi
FAIL    go.mongodb.org/atlas/test [build failed]
FAIL
```

### DIFF
--- a/test/client_test.go
+++ b/test/client_test.go
@@ -1,0 +1,28 @@
+package test
+
+import (
+	"testing"
+
+	mongodbatlas "go.mongodb.org/atlas/mongodbatlasv2"
+)
+
+func Test_Client_Config(t *testing.T) {
+	apiKey := "test"
+	apiSecret := "test"
+	host := "https://cloud.mongodb.com"
+
+	sdk, err := mongodbatlas.NewClient(
+		mongodbatlas.UseDigestAuth(apiKey, apiSecret),
+		mongodbatlas.UseBaseURL(host),
+		mongodbatlas.UseDebug(false))
+
+	if err != nil {
+		t.Error(err)
+	}
+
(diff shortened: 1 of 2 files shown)


test/client_test.go
```diff
@@ -0,0 +1,28 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	serverURL := sdk.GetConfig().Servers[0].URL
+
+	if serverURL != host {
+		t.Errorf("Expected %s, got %s", host, serverURL)
+	}
+}
```
gssbzn, line 28: let's do this an example
```suggestion
func ExampleNewClien() {
	apiKey := "test"
	apiSecret := "test"
	host := "https://cloud.mongodb.com"

	sdk, err := mongodbatlas.NewClient(
		mongodbatlas.UseDigestAuth(apiKey, apiSecret),
		mongodbatlas.UseBaseURL(host),
		mongodbatlas.UseDebug(false))

	if err != nil {
		panic(err)
	}

	serverURL := sdk.GetConfig().Servers[0].URL
	fmt.Println(serverURL)
	// Output:
	// https://cloud.mongodb.com
}
```

gssbzn (approved):
Thanks

PR #471, opened 2023-05-08
Deprecation improvements

wtrocki:
## Description

The deprecated note was extended to provide information that new version of the SDK will contain the changes. 
Only the first commit can be reviewed. Rest is regenerating the OpenAPI.
I also removed unneeded comments from the top of the files for clarity.

`Deprecated` without description is not unifined across tools so we always make sure that there is `Deprecated:` 
Ref: https://go.dev/blog/godoc#:~:text=that%20begins%20with%20%E2%80%9C-,Deprecated,-%3A%E2%80%9D%20followed%20by%20some


tools/config/go-templates/partial_header.mustache
```diff
@@ -1,18 +1,8 @@
 /*
-{{#appName}}
```
wtrocki, line 2: Our description is very very long and being part of the every file in the SDK it will contribute to the SDK file size. Removing this to make generated code look more close to what humans will write.

gssbzn (approved):
LGTM

PR #502, opened 2023-06-23
INTMDB-828: Add Schedule Parameter to Online Archive

andreaangiolillo:
## Description

This PR adds `Schedule` param to the online archive. Doc: [createOnlineArchive](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Online-Archive/operation/createOnlineArchive)

[INTMDB-828](https://jira.mongodb.org/browse/INTMDB-828)

## Type of change:

## Required Checklist:


fmenezes (dismissed):

mongodbatlas/online_archives.go
```diff
@@ -192,12 +192,24 @@ type OnlineArchive struct {
...
-	Criteria        *OnlineArchiveCriteria `json:"criteria,omitempty"`
 	DBName          string                 `json:"dbName,omitempty"`
 	GroupID         string                 `json:"groupId,omitempty"`
-	PartitionFields []*PartitionFields     `json:"partitionFields,omitempty"`
-	Paused          *bool                  `json:"paused,omitempty"`
 	State           string                 `json:"state,omitempty"`
+	Paused          *bool                  `json:"paused,omitempty"`
+	Criteria        *OnlineArchiveCriteria `json:"criteria,omitempty"`
+	Schedule        *OnlineArchiveSchedule `json:"schedule,omitempty"`
+	PartitionFields []*PartitionFields     `json:"partitionFields,omitempty"`
```
gssbzn, line 201: can you keep the fields alphabetically sorted?

mongodbatlas/online_archives.go
```diff
@@ -192,12 +192,24 @@ type OnlineArchive struct {
...
+}
+
+// OnlineArchiveSchedule represents the frequency and duration when archiving process occurs.
+type OnlineArchiveSchedule struct {
+	DayOfMonth  int32  `json:"dayOfMonth,omitempty"`
+	DayOfWeek   int32  `json:"dayOfWeek,omitempty"`
+	EndHour     int32  `json:"endHour,omitempty"`
+	EndMinute   int32  `json:"endMinute,omitempty"`
+	StartHour   int32  `json:"startHour,omitempty"`
+	StartMinute int32  `json:"startMinute,omitempty"`
```
gssbzn, line 211: do any of this allow zero(0)? omit empty will ignore it so they may need to be pointers

gssbzn: probably EndMinute needs to allow for zero, not sure if day of the week is zero=monday or not

andreaangiolillo: I didn't check it. Thanks! I will update the fields if needed

andreaangiolillo: `DayOfMonth` and `DayOfWeek` start from 1 the other can have 0. Thanks!

andreaangiolillo: updated

mongodbatlas/online_archives.go
andreaangiolillo: this order should be more memory efficient. Do you have a strong opinion here? If that is the case I will revert it. Thanks!

gssbzn: >this order should be more memory efficient

Have we enabled that linter? Personally I like alphabetical order of props as it gives you a quick way to verify if a field you are looking for exists but no strong opinion to rollback.
Now on memory efficiency I don't think these are that big and this may fall into a bit of micro improvement/premature optimization but if you want to do it I think you should enable that linter

mongodbatlas/online_archives.go
wtrocki: @andreaangiolillo LGTM
Reference: https://github.com/mongodb/atlas-sdk-go/blob/78b3662b91f20be6d8eed26f89944a36c4eb2ed3/admin/model_online_archive_schedule.go#L13

gssbzn (approved):
LGTM

wtrocki (dismissed):

mongodbatlas/online_archives.go
andreaangiolillo: > Have we enabled that linter

no, we haven't.

> Personally I like alphabetical order of props as it gives you a quick way to verify if a field you are looking for exists but no strong opinion to rollback.
> Now on memory efficiency I don't think these are that big and this may fall into a bit of micro improvement/premature optimization but if you want to do it I think you should enable that linter

Make sense. I don't think makes sense to enable the linter at this point since we will move to the autogenerated sdk. I will revert it. Thanks!

andreaangiolillo: Updated

PR #509, opened 2023-07-27
Removed omitempty from tags in ServerlessUpdateRequestParams

cveticm:
This will allow for operators to remove all tags from Serverless Instances as tags will not be omitted when sending an empty array of tags.

## Description

Please include a summary of the fix/feature/change, including any relevant motivation and context.

Link to any related issue(s): 

## Type of change:

- [x] Bug fix (non-breaking change which fixes an issue)

## Required Checklist:

## Further comments
This is a minor change to allow AKO to remove all tags from a Serverless Instance

### DIFF
--- a/mongodbatlas/serverless_instances.go
+++ b/mongodbatlas/serverless_instances.go
@@ -57,7 +57,7 @@ type ServerlessCreateRequestParams struct {
 type ServerlessUpdateRequestParams struct {
 	ServerlessBackupOptions      *ServerlessBackupOptions `json:"serverlessBackupOptions"`
 	TerminationProtectionEnabled *bool                    `json:"terminationProtectionEnabled,omitempty"`
-	Tag                          []*Tag                   `json:"tags,omitempty"`
+	Tag                          []*Tag                   `json:"tags"`
 }
 
 // ServerlessProviderSettings represents the Provider Settings of serverless instances.


wtrocki:
I think that omitempty is very confusing for arrays.
Since we have array of pointers omitEmpty acts only on array itself - not on pointers:
https://go.dev/play/p/N64Ggd7Gdth

mongodbatlas/serverless_instances.go
```diff
@@ -57,7 +57,7 @@ type ServerlessCreateRequestParams struct {
 type ServerlessUpdateRequestParams struct {
 	ServerlessBackupOptions      *ServerlessBackupOptions `json:"serverlessBackupOptions"`
 	TerminationProtectionEnabled *bool                    `json:"terminationProtectionEnabled,omitempty"`
-	Tag                          []*Tag                   `json:"tags,omitempty"`
+	Tag                          []*Tag                   `json:"tags"`
 }
```
matt-condon, line 61: Could you write a unit test for the condition you're looking to cover?

matt-condon:
> I think `omitempty` is very confusing in golang in relation to arrays (pointer values in general) Since we have array of pointers omitEmpty acts only on array itself - not on pointers: https://go.dev/play/p/N64Ggd7Gdth

@wtrocki This approach is a [convention](https://github.com/mongodb/go-client-mongodb-atlas/blob/master/mongodbatlas/accesslist_api_keys.go#L55) across this client, so I think we're fine to go with this for consistency (also any other approach would likely introduce a breaking change)

mongodbatlas/serverless_instances.go
```diff
@@ -57,7 +57,7 @@ type ServerlessCreateRequestParams struct {
 type ServerlessUpdateRequestParams struct {
 	ServerlessBackupOptions      *ServerlessBackupOptions `json:"serverlessBackupOptions"`
 	TerminationProtectionEnabled *bool                    `json:"terminationProtectionEnabled,omitempty"`
-	Tag                          []*Tag                   `json:"tags,omitempty"`
+	Tag                          []*Tag                   `json:"tags"`
```
gssbzn, line 60: this is a patch, you probably want 
```suggestion
	Tag                          *[]*Tag                   `json:"tags"`
```

Or you'll delete the tags when only updating `TerminationProtectionEnabled`

wtrocki:
> @wtrocki This approach is a [convention](https://github.com/mongodb/go-client-mongodb-atlas/blob/master/mongodbatlas/accesslist_api_keys.go#L55) across this client

I understand that. Trick is that this has no impact unless (breaking) suggestion by @gssbzn is applied. 

https://go.dev/play/p/RKmHl4b0X6p

mongodbatlas/serverless_instances.go
cveticm: Good shout, I'll make that commit and add this to tags in ServerlessCreateRequestParams and Cluster clusters.go too

wtrocki (dismissed):
LGTM after unit test covering this case and passing

mongodbatlas/serverless_instances_test.go
```diff
@@ -356,6 +357,34 @@ func TestServerlessInstances_Update(t *testing.T) {
...
+
+	serverlessInstance, _, err = client.ServerlessInstances.Update(ctx, groupID, "sample", bodyParam)
+	if err != nil {
+		t.Fatalf("ServerlessInstances.Get returned error: %v", err)
+	}
+
+	expected = &Cluster{
+		Tags: &[]*Tag{},
+	}
+
```
wtrocki, line 387: We are missing assertion above:

```
if diff := deep.Equal(serverlessInstance, expected); diff != nil {
		t.Error(diff)
	}
```

wtrocki (dismissed):

mongodbatlas/serverless_instances_test.go
```diff
@@ -359,6 +360,38 @@ func TestServerlessInstances_Update(t *testing.T) {
 	if diff := deep.Equal(serverlessInstance, expected); diff != nil {
 		t.Error(diff)
 	}
+
+	// Testing for removing tags
+	client, mux, teardown = setup()
```
gssbzn, line 365: can you please group the tests using `t.Run` for example https://github.com/mongodb/go-client-mongodb-atlas/blob/774bf48ebd3cbee3dabcda9ac3100073ca2e2af2/mongodbatlas/organizations_test.go#L28-L173

gssbzn (approved):

PR #518, opened 2023-10-30
feat: Adds acceptDataRisksAndForceReplicaSetReconfig parameter in cluster and advanced cluster

lantoli:
## Description

Adds acceptDataRisksAndForceReplicaSetReconfig parameter in cluster and advanced cluster.

Ticket: [INTMDB-1236](https://jira.mongodb.org/browse/INTMDB-1236)

## Type of change:

## Required Checklist:

### DIFF
--- a/mongodbatlas/advanced_clusters.go
+++ b/mongodbatlas/advanced_clusters.go
@@ -44,27 +44,28 @@ var _ AdvancedClustersService = &AdvancedClustersServiceOp{}
 
 // AdvancedCluster represents MongoDB cluster.
 type AdvancedCluster struct {
-	BackupEnabled                *bool                      `json:"backupEnabled,omitempty"`
-	BiConnector                  *BiConnector               `json:"biConnector,omitempty"`
-	ClusterType                  string                     `json:"clusterType,omitempty"`
-	ConnectionStrings            *ConnectionStrings         `json:"connectionStrings,omitempty"`
-	DiskSizeGB                   *float64                   `json:"diskSizeGB,omitempty"`
-	EncryptionAtRestProvider     string                     `json:"encryptionAtRestProvider,omitempty"`
-	GroupID                      string                     `json:"groupId,omitempty"`
-	ID                           string                     `json:"id,omitempty"`
-	Labels                       []Label                    `json:"labels,omitempty"`
-	MongoDBMajorVersion          string                     `json:"mongoDBMajorVersion,omitempty"`
-	MongoDBVersion               string                     `json:"mongoDBVersion,omitempty"`
-	Name                         string                     `json:"name,omitempty"`
-	Paused                       *bool                      `json:"paused,omitempty"`
-	PitEnabled                   *bool                      `json:"pitEnabled,omitempty"`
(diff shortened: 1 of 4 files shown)


mongodbatlas/advanced_clusters.go
```diff
@@ -44,27 +44,28 @@ var _ AdvancedClustersService = &AdvancedClustersServiceOp{}
 
 // AdvancedCluster represents MongoDB cluster.
 type AdvancedCluster struct {
-	BackupEnabled                *bool                      `json:"backupEnabled,omitempty"`
-	BiConnector                  *BiConnector               `json:"biConnector,omitempty"`
```
wtrocki, line 48: Can we avoid formatter changes?

wtrocki:
Check: V2 sdk contains the same field: https://github.com/mongodb/atlas-sdk-go/blob/bea1252c35376028cac30983cfb83732f4812472/docs/docs/AdvancedClusterDescription.md?plain=1#L63

wtrocki:
Good job! will approve after formatting issues are fixed

lantoli:
> Check: V2 sdk contains the same field: https://github.com/mongodb/atlas-sdk-go/blob/bea1252c35376028cac30983cfb83732f4812472/docs/docs/AdvancedClusterDescription.md?plain=1#L63

I know but we've decided in this case to wait until we migrate to new SDK to mitigate risk and fix this soon

mongodbatlas/advanced_clusters.go
lantoli: I'm not sure that can be done, it's go fmt because the new field name length is larger, I don't think we should override go fmt

wtrocki:
>  I know but we've decided in this case to wait until we migrate to new SDK to mitigate risk and fix this soon

Understood. My intention was only to ensure that the name of the property is the same and we have parity so we do not create a blocker for future migration.

mongodbatlas/advanced_clusters.go
wtrocki: >  I don't think we should override go fmt

Apologies for lack of clarity. That wasn't my intention. 

I think a good pattern in this case is to add this field as the last element. Since this field is much longer than everything else we can add comment or enter and then rest of the content is not affected:

Screenshot 2023-10-30 at 15 43 05

Hope that makes sense

wtrocki (approved):
LGTM. No blockers - formatting can be improved but not blocking

mongodbatlas/advanced_clusters.go
wtrocki: What I have done:
1. git reset --soft head~2
2. Moved field to the bottom
3. make fmt

gssbzn: In this client we like to keep fields alphabetically ordered for easy comparison in case of missing fields

lantoli:
thanks @wtrocki , I you don't mind then I'll leave it as it is because I wanted the field to be the first one exactly as in the new SDK (because they're alphabetically sorted)

wtrocki:
LGTM

